PR #49, opened 2020-12-03
[MYEN-411] update vue example

dwojno:


examples/vue-dapp/src/components/Main.vue
```diff
@@ -0,0 +1,172 @@
...
+import Vue from "vue";
+import Spinner from "./Spinner.vue";
+import axios from "axios";
+
+type Role = {
+  name: string;
+  namespace: string;
+};
+
+type User = {
```
jrhender, line 53: Seems as though the `User` type isn't being used. If so, can it be removed?

examples/vue-dapp/src/components/Main.vue
```diff
@@ -0,0 +1,172 @@
...
+    },
+    loginWithWalletConnect: async function() {
+      this.isLoading = true;
+      const { did } = await this.$IAM.initializeConnection();
+      if (did) {
+        this.did = did;
+        try {
+          await this.verifyIdentity();
+        } catch (err) {
+          this.isLoading = false;
```
jrhender, line 100: I don't think this assignment to `this.isLoading` is necessary. In all paths, won't the `this.isLoading = false` below be reached?

examples/vue-dapp/src/components/Main.vue
```diff
@@ -0,0 +1,172 @@
...
+        }
+      }
+      this.isLoading = false;
+    },
```
jrhender, line 119: Because `loginWithMetamask` and `loginWithWalletConnect` are so similar, maybe we could just pass in the initialization options and condense down to one method? Something like `login({ useMetamaskExtension}?)`

examples/vue-dapp/src/components/Main.vue
```diff
@@ -0,0 +1,172 @@
...
+      <div class="logoutContainer">
+        <button @click="logout" class="button">
+          <span>Logout</span>
+        </button>
+      </div>
+    </div>
+    <div v-else class="container">
+      <button @click="loginWithWalletConnect" class="button">
+        <img class="walletconnect" src="../assets/wallet-connect-icon.svg" />
+        <span>Login with Wallect Connect</span>
```
jrhender, line 33: Typo in Wallet

examples/vue-dapp/src/components/Main.vue
dwojno: If the function verifyIdentity will throw an error it will not go down and set `this.isLoading = false` so I need to do it in try catch

examples/vue-dapp/src/components/Main.vue
dwojno: Fixed 😄

examples/vue-dapp/src/components/Main.vue
dwojno: Fixed 😄

examples/vue-dapp/src/components/Main.vue
dwojno: Fixed 😄

manihagh (approved):

PR #52, opened 2020-12-07
[MYEN-427] update bili config to resolve nodejs issue

### LINKED ISSUE #47: Node.JS library fails to load: ES module import error
**Problem**

On server environments it is not possible to use the library due to an incorrect ESM module import (`nats.ws`). The output from `bili`, in `dist/iam-client-lib.js` line 18 is:
```
var nats_ws = require('nats.ws');
```
This causes the error:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /private/tmp/lib-test/node_modules/nats.ws/nats.mjs
```

Tested with both Node 12 and 14 LTS (`v12.20.0` and `v14.15.1`). 

**Solution**

The ESM module can be imported using `import` syntax:

```
var nats_ws = import('nats.ws');
```

My hacky workaround was to run `sed` after the `prepare` script and pack it as a local dependency in another project. Perhaps there's a way that `bili` can configure this instead. 

**Steps to reproduce**

```
cd /fresh/dir
npm init -y
npm install git+ssh://git@github.com/energywebfoundation/iam-client-lib.git#develop
npm install tslib   # seems to be necessary if not already using typescript
node
```

In the REPL:
```
> const { IAM } = require('iam-client-lib')
```

Should display the following:
```
Uncaught:
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /private/tmp/lib-test/node_modules/nats.ws/nats.mjs
    at Module.load (internal/modules/cjs/loader.js:926:11)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/private/tmp/lib-test/node_modules/iam-client-lib/dist/iam-client-lib.js:18:15)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at REPL1:1:17
    at Script.runInThisContext (vm.js:132:18)
    at REPLServer.defaultEval (repl.js:484:29)
    at bound (domain.js:430:14)
    at REPLServer.runBound [as eval] (domain.js:443:12) {
  code: 'ERR_REQUIRE_ESM'
}
```

In the same REPL, you can test the `import` syntax:
```
> const _nats = require('nats.ws')
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/adam/dev/energyweb/ocn-tools/node_modules/nats.ws/nats.mjs
> const nats = import('nats.ws')
// works
```

dwojno:
Fixes #47 

Update bili's config to bundle up the `nats.ws` library

### DIFF
--- a/bili.config.ts
+++ b/bili.config.ts
@@ -6,12 +6,17 @@ const config: Config = {
       tsconfigOverride: {
         include: ['src', 'ethers']
       }
-    }
+    },
   },
-
+  extendConfig: (config) => ({
+    ...config,
+    externals: []
+  }),
+  bundleNodeModules: ["nats.ws"],
   input: 'src/iam-client-lib.ts',
   output: {
-    format: ['cjs', 'esm']
+    format: ['cjs', 'esm'],
+    minify: true
   }
 };
 


manihagh (approved):

bili.config.ts
```diff
@@ -6,12 +6,17 @@ const config: Config = {
...
+  extendConfig: (config) => ({
+    ...config,
+    externals: []
+  }),
+  bundleNodeModules: ["nats.ws"],
   input: 'src/iam-client-lib.ts',
   output: {
-    format: ['cjs', 'esm']
+    format: ['cjs', 'esm'],
+    minify: true
```
jrhender, line 19: @dwojno Is adding minification relevant to this fix or just a good thing to do in general?

bili.config.ts
```diff
@@ -6,12 +6,17 @@ const config: Config = {
...
       }
-    }
+    },
   },
-
+  extendConfig: (config) => ({
+    ...config,
+    externals: []
+  }),
+  bundleNodeModules: ["nats.ws"],
```
jrhender, line 15: @dwojno What do you think of adding a comment here? Something like:
```
nats.ws is bundled because using standard import/require does not work for cjs format. 
nat.ws only provides a .mjs file which cannot be required.
```

bili.config.ts
dwojno: Nope, it's not relevant, but I thought that this library is being used on the browser, so the size matters 😄

bili.config.ts
```diff
@@ -6,12 +6,17 @@ const config: Config = {
       tsconfigOverride: {
         include: ['src', 'ethers']
       }
-    }
+    },
   },
-
+  extendConfig: (config) => ({
```
jrhender, line 11: @dwojno Why the need to add `extendConfig` here with the empty `externals` array? Was this needed in order to be able to use `bundleNodeModules`?

dwojno: Because the default externals are the package.json dependencies and you can only bundle node_modules that is not a dependency, so I need to clear out all externals, to bundle `nats.ws` package

PR #55, opened 2020-12-09
MYEN-409 Improvement of React client example

jrhender:
Modified the React example to be essentially the same as the current Vue example, but with React colours.
(Great job on the Vue example @dwojno , made this pretty easy 👍 )


examples/react-dapp/src/App.tsx
```diff
@@ -1,9 +1,12 @@
 import React, { useState } from 'react'
 import './App.css'
+import './Login.css'
 import { IAM, CacheServerClient } from 'iam-client-lib'
 import axios from 'axios'
-import Button from '@material-ui/core/Button'
-import { Typography } from '@material-ui/core'
```
dwojno, line 6: If you're not using material ui, please remove it from package.json 😄

examples/react-dapp/src/App.tsx
```diff
@@ -18,52 +21,129 @@ type Role = {
...
   }
-  return (
-    <div className="App">
-      <header className="App-header">
+
+  const logout = function () {
+    setDID('');
+  }
+
+  let loginJsx
```
dwojno, line 85: I would assign each view to different function like renderLogin, renderLoading, renderRoles, and main function with
`
if () {
return renderLogin()
}
if () {
return renderButton()
}
return renderRoles()
`
and run this function inside jsx

examples/react-dapp/src/App.tsx
jrhender: Nice catch, thanks, I've removed it

examples/react-dapp/src/App.tsx
```diff
@@ -18,52 +21,138 @@ type Role = {
...
+    if (isLoading) {
+      return renderLoading()
+    }
+    else if (did) {
+      return renderLoginResults();
+    }
+    else {
+      return renderLoginOptions();
+    }
+  };
```
jrhender, line 148: @dwojno Is this what you had in mind for the `loginJsx` refactor or is there a further improvement?

examples/react-dapp/src/App.tsx
jrhender: Good point, it's better to not be mutating the `loginJsx` variable

manihagh (approved):

examples/react-dapp/src/App.tsx
dwojno: Yes, perfect 👍

dwojno (approved):

PR #57, opened 2020-12-10
[MYEN-410] Angular example implementation

jrhender:
Implementation of the Angular example, it's very simpler to the Vue example.

One thing that could maybe improved (in all of the examples) to move the backend and iam-client-lib config in a single config/environment location.


examples/angular-dapp/src/app/app.component.ts
```diff
@@ -1,10 +1,59 @@
...
+
+  isLoading: boolean = false
+  did = undefined
+  roles: Role[]
+
+  async verifyIdentity() {
+    const claim = await this.iamService.iam.createIdentityProof();
+    const {
+      data: { token }
+    } = await axios.post<{ token: string }>("https://did-auth-demo.energyweb.org/login", {
```
manihagh, line 24: It's okay for now but I would recommend to use most commonly used HTTPClient from @angular/common/http

manihagh (approved):

examples/angular-dapp/src/app/app.component.ts
jrhender: Thanks for the feedback, I will merge for now, and then fix at later point

PR #58, opened 2020-12-10
[MYEN-385] Handling login error in client examples

jrhender:
Catching login errors in all three examples and displaying standard error message.

Message is:

```
Error occured with login.
If you rejected the signing requests, please try again and accept.
If this is your first time logging in, your account needs a small amount of Volta token to create a DID Document.
A Volta token can be obtained from the <a href="https://voltafaucet.energyweb.org/">Volta Faucet</a>.
```

Is it reasonable that the message is Volta specific?

Not displaying actually message from error as it is not always meaningful (e.g. when rejecting Metamask signing request, the `err.message` that is caught is `"[object Object]"`)


manihagh (approved):
Great

PR #77, opened 2020-12-21
Fixing isBrowser logic and adding tests

jrhender:
I think the logic of `isBrowser` might have been backwards. I added some tests and fixed. Let me know if this makes sense

### DIFF
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,8 +1,8 @@
-export const isBrowser = (processObj: processProps = process) =>
-  typeof processObj === "undefined" || processObj.release?.name.search(/node|io.js/) === -1;
-
-interface processProps {
-    release?: {
-        name: string
-    }
-}
+/**
+ * isBrowser creates a new function using a new Function() constructor.
+ * This is so the execution scope of this is binded to the global scope.
+ * We can compare the global scope directly to the expected value of "window".
+ * An alternative approach of checking a javascript variable (e.g. window or process)
+ * is fallible because javascript variables can be redefined by inner scopes.
+ */
+export const isBrowser = new Function("try {return this===window;} catch(e) {return false;}");
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,2 +1,8 @@
-export const isBrowser = () =>
-  typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1;
+export const isBrowser = (processObj: processProps = process) =>
+  typeof processObj === "undefined" || processObj.release?.name.search(/node|io.js/) === -1;
+
+interface processProps {
+    release?: {
+        name: string
+    }
+}
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,8 +1,8 @@
 /**
- * isBrowser creates a new function using a new Function() constructor.
- * This is so the execution scope of this is binded to the global scope.
- * We can compare the global scope directly to the expected value of "window".
- * An alternative approach of checking a javascript variable (e.g. window or process)
- * is fallible because javascript variables can be redefined by inner scopes.
+ * Though the process variable can be redefined, this approach is suitable for now.
+ * 
+ * Another option is to use new Function() constructor so the execution scope of this is binded to the global scope.
+ * However, this fails if Content-Security-Policy (CSP) script-src does not allow 'unsafe-eval'.
  */
-export const isBrowser = new Function("try {return this===window;} catch(e) {return false;}");
+export const isBrowser = () =>
+  typeof process === "undefined" || process.release?.name.search(/node|io.js/) === -1;
--- a/test/utils/isBrowser.test.ts
+++ b/test/utils/isBrowser.test.ts
@@ -1,33 +1,8 @@
-import { isBrowser } from '../../src/utils/isBrowser'
+import { isBrowser } from '../../src/utils/isBrowser';
 
 describe("isBrowser Test", () => {
-    it('returns false if process release name is node', () => {
-        const mockProcess = {
-            release: {
-                name: 'node'
-            }
-        };
-        const result = isBrowser(mockProcess);
+    it('returns false if running in node', () => {
+        const result = isBrowser();
         expect(result).toBeFalsy();
-    });
-
-    it('returns false if process release name is io.js', () => {
-        const mockProcess = {
-            release: {
-                name: 'io.js'
-            }
-        };
-        const result = isBrowser(mockProcess);
-        expect(result).toBeFalsy();
-    });
-
-    it('returns true if process release name is not node', () => {
-        const mockProcess = {
-            release: {
-                name: 'notNode'
-            }
-        };
-        const result = isBrowser(mockProcess);
-        expect(result).toBeTruthy();
     });
 });
--- a/test/utils/isBrowser.test.ts
+++ b/test/utils/isBrowser.test.ts
@@ -1,0 +1,33 @@
+import { isBrowser } from '../../src/utils/isBrowser'
+
+describe("isBrowser Test", () => {
+    it('returns false if process release name is node', () => {
+        const mockProcess = {
+            release: {
+                name: 'node'
+            }
+        };
+        const result = isBrowser(mockProcess);
+        expect(result).toBeFalsy();
+    });
+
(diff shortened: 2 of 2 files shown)


jrhender:
@JGiter I am confused however how the `isRunningBrowser` checks worked if the `isBrowser` logic was indeed incorrect 🤔 
Does this make sense to you?

JGiter (commented):
defining the environment by checking `process` variable is not reliable because this var can be overridden. Better to check the global variable in a function that will be executed in the global scope:

```
const isBrowser = new Function ("try {return this === window;} catch (e) {return false;}");
if (isBrowser ()) console.log ("running under browser");
```
Also there is a npm package `is-browser`

JGiter:
> @JGiter I am confused however how the `isRunningBrowser` checks worked if the `isBrowser` logic was indeed incorrect 
> Does this make sense to you?

Probably webpack adds process object to the build
image

jrhender:
@JGiter So as I understand it, the `isBrowser` logic (`typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1`) in the deployed `iam-client-lib` gives the opposite result of what I would expect. In other words, if `process.release.name === "node"`, then above logic returns true.

This code for instance runs without error for me in Node.js using the `iam-client-lib` npm package:
```
import { IAM } from 'iam-client-lib'

(async () => {
    const iam = new IAM({
        privateKey: 'bafe62d65126513025b683a5a78b48c7ad5b4cea5661cfc9e5463b192ce3d408',
        rpcUrl: "https://volta-rpc.energyweb.org",
        chainId: 73799,
    })
    await iam.initializeConnection()
    await iam.createSelfSignedClaim({ "data": {test: "some data"}} )
})();
```

Expected behaviour would be that above code fails because `createSelfSignedClaim` has a check that it is running in browser:
```
  async createSelfSignedClaim({ data }: { data: Record<string, unknown> }) {
    if (!this._runningInBrowser) {
      throw new MethodNotAvailableInNodeEnvError("createSelfSignedClaim");
    }
```

**So what has me confused** is that were saying in Monday's standup that you ran into `MethodNotAvailableInNodeEnvErrors` when writing tests? And so I don't understand how you would have the errors, assuming the `isBrowser` logic is giving "opposite results".

jrhender:
I think that the incorrect `isBrowser` logic isn't affecting Switchboard because the methods have `MethodNotAvailableInNodeEnvErrors` aren't being used in Switchboard:
`updateDidDocument` function -> not used in Switchboard
`revokeDidDocument` function -> not used in Switchboard

> > @JGiter I am confused however how the `isRunningBrowser` checks worked if the `isBrowser` logic was indeed incorrect
> > Does this make sense to you?
> 
> Probably webpack adds process object to the build
> image

Sorry, how did you get process to print out in the console? I don't see it when I try to access it: 
image

JGiter:
> @JGiter So as I understand it, the `isBrowser` logic (`typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1`) in the deployed `iam-client-lib` gives the opposite result of what I would expect. In other words, if `process.release.name === "node"`, then above logic returns true.
> 
> This code for instance runs without error for me in Node.js using the `iam-client-lib` npm package:
> 
> ```
> import { IAM } from 'iam-client-lib'
> 
> (async () => {
>     const iam = new IAM({
>         privateKey: 'bafe62d65126513025b683a5a78b48c7ad5b4cea5661cfc9e5463b192ce3d408',
>         rpcUrl: "https://volta-rpc.energyweb.org",
>         chainId: 73799,
>     })
>     await iam.initializeConnection()
>     await iam.createSelfSignedClaim({ "data": {test: "some data"}} )
> })();
> ```
> 
> Expected behaviour would be that above code fails because `createSelfSignedClaim` has a check that it is running in browser:
> 
> ```
>   async createSelfSignedClaim({ data }: { data: Record<string, unknown> }) {
>     if (!this._runningInBrowser) {
>       throw new MethodNotAvailableInNodeEnvError("createSelfSignedClaim");
>     }
> ```
> 
> So what has me confused is that you were saying in Monday's standup that you ran into `MethodNotAvailableInNodeEnvErrors` when writing tests? And so I don't understand how you would have the errors, assuming the `isBrowser` logic is giving "opposite results".

JGiter:
> > @JGiter So as I understand it, the `isBrowser` logic (`typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1`) in the deployed `iam-client-lib` gives the opposite result of what I would expect. In other words, if `process.release.name === "node"`, then above logic returns true.
> > This code for instance runs without error for me in Node.js using the `iam-client-lib` npm package:
> > ```
> > import { IAM } from 'iam-client-lib'
> > 
> > (async () => {
> >     const iam = new IAM({
> >         privateKey: 'bafe62d65126513025b683a5a78b48c7ad5b4cea5661cfc9e5463b192ce3d408',
> >         rpcUrl: "https://volta-rpc.energyweb.org",
> >         chainId: 73799,
> >     })
> >     await iam.initializeConnection()
> >     await iam.createSelfSignedClaim({ "data": {test: "some data"}} )
> > })();
> > ```
> > 
> > 
> > Expected behaviour would be that above code fails because `createSelfSignedClaim` has a check that it is running in browser:
> > ```
> >   async createSelfSignedClaim({ data }: { data: Record<string, unknown> }) {
> >     if (!this._runningInBrowser) {
> >       throw new MethodNotAvailableInNodeEnvError("createSelfSignedClaim");
> >     }
> > ```
> > 
> > 
> > So what has me confused is that you were saying in Monday's standup that you ran into `MethodNotAvailableInNodeEnvErrors` when writing tests? And so I don't understand how you would have the errors, assuming the `isBrowser` logic is giving "opposite results".

@jrhender You are right. Logic is incorrect. But in my case, the reason was that iam-client-lib used an old version of ew-did-registry, which led to failing of iam initialization and  createSubdomain throws MethodNotAvailableInNodeEnvError if some of the iam fields are not initialized

JGiter:
> > > @JGiter So as I understand it, the `isBrowser` logic (`typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1`) in the deployed `iam-client-lib` gives the opposite result of what I would expect. In other words, if `process.release.name === "node"`, then above logic returns true.
> > > This code for instance runs without error for me in Node.js using the `iam-client-lib` npm package:
> > > ```
> > > import { IAM } from 'iam-client-lib'
> > > 
> > > (async () => {
> > >     const iam = new IAM({
> > >         privateKey: 'bafe62d65126513025b683a5a78b48c7ad5b4cea5661cfc9e5463b192ce3d408',
> > >         rpcUrl: "https://volta-rpc.energyweb.org",
> > >         chainId: 73799,
> > >     })
> > >     await iam.initializeConnection()
> > >     await iam.createSelfSignedClaim({ "data": {test: "some data"}} )
> > > })();
> > > ```
> > > 
> > > 
> > > Expected behaviour would be that above code fails because `createSelfSignedClaim` has a check that it is running in browser:
> > > ```
> > >   async createSelfSignedClaim({ data }: { data: Record<string, unknown> }) {
> > >     if (!this._runningInBrowser) {
> > >       throw new MethodNotAvailableInNodeEnvError("createSelfSignedClaim");
> > >     }
> > > ```
> > > 
> > > 
> > > So what has me confused is that you were saying in Monday's standup that you ran into `MethodNotAvailableInNodeEnvErrors` when writing tests? And so I don't understand how you would have the errors, assuming the `isBrowser` logic is giving "opposite results".
> 
> @jrhender You are right. Logic is incorrect. But in my case, the reason was that iam-client-lib used an old version of ew-did-registry, which led to failing of iam initialization and createSubdomain throws MethodNotAvailableInNodeEnvError if some of the iam fields are not initialized
Also I don't quite understand why the use of the library is limited by the browser? From the code I can see that MethodNotAvailableInNodeEnvError is thrown if _ensRegistry or _signer is not initialized. But they do not have to be initialized in the browser. In addition, this limitation makes testing difficult because I need somehow to simulate the browser

jrhender:
@JGiter Took your advice and now using `Function` constructor to check the global scope, thanks.

jrhender:
@JGiter nevermind, I went back to just checking the `process` variable because Function constructor approach fails if unsafe-eval is not allowed in Content Security Policy script-src directory

(below is a screenshot of trying it out in the Fastmail web app, which I just happened to have open)

image

JGiter (approved):

PR #78, opened 2020-12-23
feature/tests

JGiter:
- [x] Setting up test suite
- [x] Test root node creation
- [x] Test organization creation
- [x] Test application creation
- [x] Test sub-organization creation
- [x] Test role creation
- [x] Test delete application 

Tests of ownership transfer will be added after simplifying node transfer 
Version of the `nats.ws` is pinned because newest versions require reconfiguration of `babel`, for example including class properties plugin
`process.release.name` set to empty string during tests and therefore isBrowser test is changed


package.json
```diff
@@ -91,20 +65,30 @@
...
+    "@types/chai": "^4.2.14",
+    "@types/chai-as-promised": "^7.1.3",
     "@types/jest": "^26.0.13",
     "@types/lodash.difference": "^4.5.6",
     "@types/uuid": "^8.3.0",
     "@typescript-eslint/eslint-plugin": "^4.0.1",
     "@typescript-eslint/parser": "^4.0.1",
+    "babel-jest": "^26.6.3",
     "bili": "^5.0.5",
+    "chai": "^4.2.0",
```
dwojno, line 81: While using `jest` you don't need chai anymore :) `jest` has built in assertion (expect), mocks etc :)

JGiter: Ok. thanks. will remove it

test/env.js
```diff
@@ -0,0 +1,18 @@
+const Environment = require('jest-environment-jsdom');
+const nodeCrypto = require('crypto');
+
+module.exports = class extends Environment {
+    async setup() {
+        await super.setup();
+        if (typeof this.global.TextEncoder === 'undefined') {
```
jrhender, line 7: Personally, I think a comment would be helpful to explain why this code is needed. For instance, why is the conditional statement necessary? Is it not known ahead of time whether or not we need to set these global values? I don't see `this.global` used elsewhere in the tests.

README.md
```diff
@@ -78,7 +67,26 @@ export class Sample {
 
 ```
 
+## Development
```
jrhender, line 70: I think instructions on how to run the tests would be helpful

jrhender:
@JGiter One of the tests is failing when I run `npm run test:prod`. Should it be passing?

image

test/env.js
JGiter: `TextEncoder` and `TextDecoder` classes are needed by some iam-client-lib dependency and are provided by the browser environment, while in the Node these classes are contained in `util` package. Because of this I add them to globals. You are right that the purpose of this setup is unclear. I will add doc comments

README.md
JGiter: > @JGiter One of the tests is failing when I run `npm run test:prod`. Should it be passing?
> 
> image

This is strange. I will look at it. Thanks for noticing.

JGiter: @jrhender I can't reproduce this error and `test:prod` is used during CI, which is passes

jrhender: @JGiter I think that the issue is that the `getSubdomains` function is returning instead of continuing if same/less length. Shouldn't line 1045 in the screenshot below be `if (nameArray.length <= role.length) continue;`? The test passes for me if I change it to that. What is happening with `return` is that the function is returning `undefined` because `role` (`['org1', 'root']`) is the same length as `nameArray` (`['root', 'org1']`).
image

JGiter: This error is fixed in this PR

jrhender: Ah yes, my fault, I just realized that. Thanks

jrhender (approved):

package.json
```diff
@@ -43,43 +43,17 @@
...
     "@ensdomains/ens": "^0.4.5",
     "@ensdomains/resolver": "^0.2.4",
-    "@ew-did-registry/claims": "0.0.1-alpha.776.0",
-    "@ew-did-registry/did": "0.0.1-alpha.632.0",
-    "@ew-did-registry/did-document": "0.0.1-alpha.776.0",
-    "@ew-did-registry/did-ethr-resolver": "0.0.1-alpha.776.0",
-    "@ew-did-registry/did-resolver-interface": "0.0.1-alpha.776.0",
-    "@ew-did-registry/jwt": "0.0.1-alpha.676.0",
-    "@ew-did-registry/keys": "0.0.1-alpha.632.0",
+    "@ew-did-registry/claims": "prerelease",
```
dwojno, line 50: Maybe we should stick to static versions of `did-library`, cause always getting the newest version while building may break something

package.json
```diff
@@ -91,20 +65,28 @@
     "uuid": "^7.0.3"
   },
   "devDependencies": {
+    "@babel/core": "^7.12.10",
+    "@babel/plugin-transform-regenerator": "^7.12.1",
+    "@babel/preset-env": "^7.12.11",
     "@typechain/ethers-v4": "^1.0.1",
+    "@types/chai": "^4.2.14",
```
dwojno, line 72: I think this 2 packages are not needed now :)

package.json
JGiter: I agree. I also had to pin the version of the `nats.ws` because the new version uses class properties which is not yet supported by `babel`.

package.json
JGiter: Thanks for noticing. I removed types/chai, but why we don't need @typechain/ethers-v4? Interfaces that it generates are used in lib

dwojno: Sorry, I was not clear, I ment chai and chai as promised types :)

PR #81, opened 2021-01-05
[MYEN-510] add method to reject claim

dwojno:


JGiter (approved):

src/iam.ts
```diff
@@ -53,11 +53,12 @@ type InitializeData = {
 
 export interface IMessage {
   id: string;
-  token: string;
+  token?: string;
   issuedToken?: string;
   requester: string;
   claimIssuer: string[];
   acceptedBy?: string;
+  isRejected?: boolean;
```
jrhender, line 61: I would suggest documenting the meaning of the properties of `IMessage`, in particular that `isRejected` being false or undefined (and with an issuedToken) implies that the claim is accepted/issued.

Or perhaps even better would be to have different interfaces for the different types of messages: `IClaimRequest`, `IClaimIssuance`, `IClaimRejection`. And then the different methods in `cacheServerClient` could accept the right interface. This would make the different acceptable messages clear and avoid the nullable properties.

jrhender (commented):
I think it could definitely work as is, though I think it's a bit confusing that `IMessage` is beginning to have so many nullable properties.

src/iam.ts
dwojno: Yeah, great idea, I've changed the code, please check if this is ok 😄

jrhender (approved):
Looks great 🚀

PR #83, opened 2021-01-05
Revert "Fixing isBrowser logic and adding tests"

jrhender:
Reverts energywebfoundation/iam-client-lib#77 

I found major issues testing the `isBrowser` fix again locally. My initial testing was incorrect.

### DIFF
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,2 +1,4 @@
-export const isBrowser = () =>
-  typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1;
+// We were previously checking if were inBrowser because there was no way to setup Signer in node.
+// However, we can now set the signer by passing in the privateKey and so this check isn't needed.
+// TODO: Remove references to this function. One thing to check is that NATS works in node.
+export const isBrowser = () => true;
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,4 +1,2 @@
-// We were previously checking if were inBrowser because there was no way to setup Signer in node.
-// However, we can now set the signer by passing in the privateKey and so this check isn't needed.
-// TODO: Remove references to this function. One thing to check is that NATS works in node.
-export const isBrowser = () => true;
+export const isBrowser = () =>
+  typeof process?.release?.name === "undefined" || process.release.name.search(/node|io.js/) === -1;
--- a/src/utils/isBrowser.ts
+++ b/src/utils/isBrowser.ts
@@ -1,8 +1,2 @@
-/**
- * Though the process variable can be redefined, this approach is suitable for now.
- * 
- * Another option is to use new Function() constructor so the execution scope of this is binded to the global scope.
- * However, this fails if Content-Security-Policy (CSP) script-src does not allow 'unsafe-eval'.
- */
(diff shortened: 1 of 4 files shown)


src/utils/isBrowser.ts
```diff
@@ -1,8 +1,2 @@
-/**
- * Though the process variable can be redefined, this approach is suitable for now.
- * 
- * Another option is to use new Function() constructor so the execution scope of this is binded to the global scope.
- * However, this fails if Content-Security-Policy (CSP) script-src does not allow 'unsafe-eval'.
- */
 export const isBrowser = () =>
-  typeof process === "undefined" || process.release?.name.search(/node|io.js/) === -1;
+  typeof process !== "undefined" && process.release?.name.search(/node|io.js/) !== -1;
```
JGiter, line 2: What is wrong with this approach?

jrhender: @JGiter In the Switchboard app, the `process` variable is defined, but `process.release` is not, so the new `isBrowser` is returning `false` in Switchboard 🤦
image

JGiter (approved):

jrhender:
Based on conversation in standup today, `isBrowser` check was only needed before we had privateKey Signer method. We should be able to remove now. This is a temporary fix so that `iam-client-lib` passes tests and works in Switchboard.

src/utils/isBrowser.ts
```diff
@@ -1,8 +1,4 @@
...
- * 
- * Another option is to use new Function() constructor so the execution scope of this is binded to the global scope.
- * However, this fails if Content-Security-Policy (CSP) script-src does not allow 'unsafe-eval'.
- */
-export const isBrowser = () =>
-  typeof process === "undefined" || process.release?.name.search(/node|io.js/) === -1;
+// We were previously checking if were inBrowser because there was no way to setup Signer in node.
+// However, we can now set the signer by passing in the privateKey and so this check isn't needed.
+// TODO: Remove references to this function. One thing to check is that NATS works in node.
+export const isBrowser = () => true;
```
JGiter, line 4: @jrhender Wouldn't it be better to exclude isBrowser check at all?

jrhender: @JGiter I left in `isBrowser` because I believe NATS will only work in browser.

jrhender:
I wasn't able to test `isBrowser` returning `false`.

For instance, this test fails, even though it is using the `node` jest-environment. Should we create our own jest test environment where `isBrowser` returns false so that we can test what behaviour would be in `node`?

```
/**
 * @jest-environment node
 */

describe("isBrowser.Node Test", () => {
  it('returns false if running in node', () => {
    const result = isBrowser();
    expect(result).toBe(false);
  });
});
```

JGiter:
> I wasn't able to test `isBrowser` returning `false`.
> 
> For instance, this test fails, even though it is using the `node` jest-environment. Should we create our own jest test environment where `isBrowser` returns false so that we can test what behaviour would be in `node`?
> 
> ```
> /**
>  * @jest-environment node
>  */
> 
> describe("isBrowser.Node Test", () => {
>   it('returns false if running in node', () => {
>     const result = isBrowser();
>     expect(result).toBe(false);
>   });
> });
> ```

@jrhender This test fails because to test IAM I added `process.release.name = ''` in test environment (test/env.js)

PR #84, opened 2021-01-06
[MYEN-515] simplify ownership transfer

JGiter:
iam-client-lib does not allow transferring ownership of the top-level node that you own, because transferring the node requires ownership of the parent node. This PR proposes to remove the parent node ownership restriction
- [x] Exclude parent organization from ownership validation and transferring of application

To discuss: these changes allows owner of application not to be owner of app parent organization. Is this consistent with business requirements? Should similar approach be implemented for organization and role transferring?


jrhender (approved):
Looks good to me. I don't see an issue with having the owner of an app/role/org not owning the parent namespace.

PR #87, opened 2021-01-12
[MYEN-497] update search implementation

dwojno:


src/cacheServerClient/cacheServerClient.ts
```diff
@@ -96,20 +101,24 @@ export class CacheServerClient implements ICacheServerClient {
...
+  async getNamespaceBySearchPhrase({
+    types,
+    search
+  }: {
+    types?: ENSNamespaceTypes[];
+    search: string;
+  }) {
+    if (types && types.length > 0) {
+      const searchTypes = types.map(type => type.charAt(0).toUpperCase() + type.slice(1));
+      const { data } = await this.httpClient.get<IOrganization[]>(`/namespace/search/${search}`, {
```
jrhender, line 113: I'm a little confused. Maybe it doesn't matter as data is being operated on but this `httpClient` call always returns `IOrganization[]` even though the function itself could return `IOrganization[] | IApp[] | IRole[]` based on `ICacheServerClient`. Should the type in the angle brackets be `IOrganization[] | IApp[] | IRole[]`?

dwojno: Yeah, sorry my mistake I forgot about this :)

dwojno: Fixed :)

src/cacheServerClient/cacheServerClient.ts
```diff
@@ -96,20 +100,28 @@ export class CacheServerClient implements ICacheServerClient {
...
+            types
+          },
+          paramsSerializer: params => {
+            return stringify(params, { arrayFormat: "brackets" });
+          }
+        }
+      );
+      return data;
+    }
+    const { data } = await this.httpClient.get<IOrganization[]>(`/namespace/search/${search}`);
```
jrhender, line 124: Maybe still one spot that needs `IOrganization[] | IApp[] | IRole[]`. Regardless, PR can be merged as is or after it's fixed.

jrhender (approved):
👍

PR #92, opened 2021-01-19
[MYEN-547] add connected network checking

dwojno:


src/iam.ts
```diff
@@ -83,8 +82,11 @@ export const NATS_EXCHANGE_TOPIC = "claim.exchange";
  */
 export class IAM extends IAMBase {
   static async isMetamaskExtensionPresent() {
-    const provider = await detectEthereumProvider({ mustBeMetaMask: true });
-    return !!provider;
+    const provider = (await detectEthereumProvider({ mustBeMetaMask: true })) as {
+      chainId: string;
+    };
+    const metamaskChainId = parseInt(provider.chainId, 16);
+    return metamaskChainId === VOLTA_CHAIN_ID && !!provider;
```
JGiter, line 89: If metamaskChainId then provider is not undefined

JGiter (changes requested):

src/iam.ts
dwojno: 👍

JGiter (approved):

PR #93, opened 2021-01-19
Myen 549 keymanager integration

jrhender:
- [x] Providing KeyManager WalletConnect option which directly opens KM site in new window/tab.
- [x] Using WalletProvider instead of useMetamask and useEwKM. This is more extensible in the event of future type and keeps selection mutually exclusive
- [x] Can pass in option KM url into `IAM` constructor
- [x] Deprecate `useMetaMask` option in favor of `WalletProvider` enum
- [x] Fix & refactor of `getPublicKey` function. See #100 for details 

**Note**: This PR could be merged in order to deploy to dev environment to test but currently KM login is failing as retrieval of publicKey from signature does not validate

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -43,6 +43,7 @@ import {
   IRoleDefinition
 } from "./cacheServerClient/cacheServerClient.types";
 import detectEthereumProvider from "@metamask/detect-provider";
+import { WalletProvider } from "./types/WalletProvider";
 
 type InitializeData = {
   did: string | undefined;
@@ -122,8 +123,8 @@ export class IAM extends IAMBase {
     {
       useMetamaskExtension,
       reinitializeMetamask,
-      useEwKeyManager
-    }: { useMetamaskExtension: boolean; reinitializeMetamask?: boolean, useEwKeyManager?: boolean } = {
+      walletProvider
+    }: { useMetamaskExtension?: boolean; reinitializeMetamask?: boolean, walletProvider?: WalletProvider } = {
       useMetamaskExtension: false
     }
   ): Promise<InitializeData> {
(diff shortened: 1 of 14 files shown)


src/iam.ts
```diff
@@ -125,15 +126,24 @@ export class IAM extends IAMBase {
...
-      useMetamaskExtension: false
+      reinitializeMetamask,
+      walletProvider
+    }: { useMetamaskExtension?: boolean; reinitializeMetamask?: boolean, walletProvider?: WalletProvider } = {
     }
   ): Promise<InitializeData> {
     try {
+      // Deprecated warning below can be removed after param is removed
+      if (useMetamaskExtension !== undefined) {
+        console.warn("Use of useMetamaskExtension parameter is deprecated. Use walletProvider enumeration instead.")
```
jrhender, line 137: @dwojno @manihagh Do you agree with this approach of signalling that a param is deprecated to users of `iam-client-lib`

dwojno: @jrhender I think we should just switch to new approach using walletProvider enum, as for now only switchboard is using iam-client-lib so it should be fine. I think leaving it both may cause a lot of mess in initializing connection methods on iam-client-lib, what do you think?

jrhender: @dwojno Thanks for the advice, I removed the `useMetamaskExtension` param

src/iam/iam-base.ts
```diff
@@ -262,31 +273,58 @@ export class IAMBase {
...
-        bridge: this._connectionOptions.bridgeUrl
+        bridge: this._connectionOptions.bridgeUrl,
+        qrcode: showQRCode
       });
+
+      if (walletProvider === WalletProvider.EwKeyManager) {
+        this._walletConnectProvider.wc.on("display_uri", (err, payload) => {
+          // uri is expected to be WalletConnect Standard URI https://eips.ethereum.org/EIPS/eip-1328
+          const wcUri = payload.params[0];
+          console.log(wcUri);
```
dwojno, line 293: If not needed please remove console.log 😄

jrhender: Thanks, good catch, I've removed it.

dwojno (approved):

PR #95, opened 2021-01-22
[MYEN-373] add login in to cache server

dwojno:
- [x] iam-client-lib logges in to iam-cache-server
- [x] iam-client-lib request data using authentication with JWT token
- [x] iam-client-lib asks for only one signing message to get public key and generate identity proof
- [x] refactor auth module
- [x] iam-client-lib handles non authorized errors

### DIFF
--- a/src/cacheServerClient/cacheServerClient.ts
+++ b/src/cacheServerClient/cacheServerClient.ts
@@ -13,6 +13,7 @@ import {
 import { IClaimIssuance, IClaimRejection, IClaimRequest } from "../iam";
 import { IDIDDocument } from "@ew-did-registry/did-resolver-interface";
 export interface ICacheServerClient {
+  login: (claim: string) => Promise<void>;
   getRoleDefinition: ({ namespace }: { namespace: string }) => Promise<IRoleDefinition>;
   getOrgDefinition: ({ namespace }: { namespace: string }) => Promise<IOrganizationDefinition>;
   getAppDefinition: ({ namespace }: { namespace: string }) => Promise<IAppDefinition>;
@@ -74,6 +75,11 @@ export class CacheServerClient implements ICacheServerClient {
     this.httpClient = axios.create({
       baseURL: url
     });
+  }
+
+  async login(claim: string) {
+    const { data } = await this.httpClient.post<{ token: string }>("/login", { claim });
+    this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${data.token}`;
   }
 
   async getRoleDefinition({ namespace }: { namespace: string }) {
--- a/src/cacheServerClient/cacheServerClient.ts
+++ b/src/cacheServerClient/cacheServerClient.ts
@@ -77,8 +77,8 @@ export class CacheServerClient implements ICacheServerClient {
     });
   }
 
-  async login(claim: string) {
-    const { data } = await this.httpClient.post<{ token: string }>("/login", { claim });
+  async login(identityToken: string) {
+    const { data } = await this.httpClient.post<{ token: string }>("/login", { identityToken });
     this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${data.token}`;
   }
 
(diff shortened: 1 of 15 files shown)


src/iam/iam-base.ts
```diff
@@ -199,98 +195,127 @@ export class IAMBase {
     this.setupENS();
   }
 
-  private async setupProvider({
+  private async initSigner({
```
jrhender, line 198: Nice rename 👍

src/iam/iam-base.ts
```diff
@@ -199,98 +195,127 @@ export class IAMBase {
...
-        },
-        infuraId: this._connectionOptions.infuraId,
-        bridge: this._connectionOptions.bridgeUrl
-      });
-      await this._walletConnectProvider.enable();
-      this._signer = new providers.Web3Provider(this._walletConnectProvider).getSigner();
-      const publicKey = await this.getPublicKey();
-      this._didSigner = new IdentityOwner(this._signer, this._provider, publicKey);
+      const provider = new providers.Web3Provider(metamaskProvider);
+      this._signer = provider.getSigner();
```
jrhender, line 250: I find this part a bit confusing as I think the outcome is that `this._signer` is associated with the Web3Provider but then we pass a different (JsonRpc) provider into the `IdentityOwner` constructor. So `IdentityOwner` is then kind of associated with two different providers 🤔 ? I think that could be a source of confusion as to which network/chain "Switchboard" is actually connected to. Not super related to this PR and I don't have a great solution, just bringing it up for thought 😅

src/iam/iam-base.ts
```diff
@@ -199,98 +195,127 @@ export class IAMBase {
...
+        iss: `did:ethr:${address}`,
+        claimData: {
+          blockNumber: await this._provider?.getBlockNumber()
+        }
+      };
+
+      const encodedPayload = base64url(JSON.stringify(payload));
+      return { encodedHeader, encodedPayload };
+    }
+    throw Error("Signer not initialized");
```
jrhender, line 286: I would use the same `ERROR_MESSAGES.SIGNER_NOT_INITIALIZED` here

src/iam/iam-base.ts
```diff
@@ -199,98 +195,127 @@ export class IAMBase {
...
     }
-    if (this._runningInBrowser) {
-      const metamaskProvider = (await detectMetamask({
-        mustBeMetaMask: true
-      })) as any;
-      if (metamaskProvider && useMetamask) {
-        if (initializeMetamask) {
-          await metamaskProvider.request({
+    if (!this._runningInBrowser) {
+      throw new Error("Initialization of signer not possible in nodejs env without private key ");
```
jrhender, line 213: I would add a new entry to `ERROR_MESSAGES` to use here.

jrhender (approved):
A couple comments to add to use `ERROR_MESSAGES` but otherwise looks good.

src/cacheServerClient/cacheServerClient.ts
```diff
@@ -76,6 +77,11 @@ export class CacheServerClient implements ICacheServerClient {
     });
   }
 
+  async login(claim: string) {
+    const { data } = await this.httpClient.post<{ token: string }>("/login", { claim });
+    this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${data.token}`;
```
jrhender, line 82: Another aspect I thought of is whether or not this is the way that the jwt should be stored. I think this is safest spot as it in browser memory (safer than localStorage or cookie, but has downside that don't persist on new tab or refresh). Is this your understanding as well @dwojno @manihagh ?

dwojno: @jrhender yes, I agree we should add option to store jwt in the cookie as configured option, but still we need to have an option to return token for nodejs apps to be able to login

src/iam/iam-base.ts
dwojno: 👍

src/iam/iam-base.ts
dwojno: 👍

src/cacheServerClient/cacheServerClient.ts
jrhender: @dwojno Ah, sorry, I wasn't trying to say that we necessarily need an option to store jwt as a cookie. I think the way that it is currently being done in this PR is better than cookie or localStorage for our requirements. I just wanted to confirm that this was your and Mani's view as well.

PR #98, opened 2021-01-26
fix: update logic for changing ownership and deleting

dwojno:
Before merging [this iam-cache-server PR](https://github.com/energywebfoundation/iam-cache-server/pull/58) needs to be merged


JGiter (approved):

src/iam/iam-base.ts
```diff
@@ -523,6 +523,27 @@ export class IAMBase {
...
+        emptyAddress,
+        ttl,
+        this._transactionOverrides
+      );
+
+      await tx.wait();
+
+      return;
+    }
+    throw new MethodNotAvailableInNodeEnvError("deleteSubdomain");
```
jrhender, line 544: Just curious, why is this method not available in node?

jrhender (approved):
Looks good, just a comment about error type

src/iam/iam-base.ts
dwojno: Yeah, good point 👍

PR #100, opened 2021-01-27
fix(getPublicKey): check address when getPublicKey

jrhender:
The gist of this change is that:
- Only sending the `hash`, not the `digest` to signer to sign. The expectation is that the signer will add the `"\x19Ethereum Signed Message:\n" + len(message)` prefix
- When recovering the publicKey, make sure that the correct message is used (and hence the correct public key recovered) by validating against the address
- In the event that the wallet isn't adding the prefix, we can still attempt to recover the public key


JGiter (approved):

src/iam/iam-base.ts
```diff
@@ -306,17 +306,25 @@ export class IAMBase {
...
+      const sig = await this._signer.signMessage(arrayify(hash));
+
+      // Computation of the digest in order to recover the public key under the
+      // assumption that signer was made as per the eth_sign spec (https://eth.wiki/json-rpc/API#eth_sign)
       const digest = hashMessage(arrayify(hash));
-      if (((this._signer?.provider as any)?.provider as WalletConnectProvider)?.isWalletConnect) {
-        const sig = await this._signer.signMessage(arrayify(digest));
-        return computePublicKey(recoverPublicKey(digest, sig), true).slice(2);
+      const publicKey_digest = recoverPublicKey(digest, sig);
+      if (address === computeAddress(publicKey_digest)) {
```
adamstaveley, line 319: You might run into issues comparing non-checksum with checksum addresses. See https://docs.ethers.io/v4/notes.html#checksum-address.

```suggestion
      if (getAddress(address) === getAddress(computeAddress(publicKey_digest))) {
```

src/iam/iam-base.ts
```diff
@@ -306,17 +306,25 @@ export class IAMBase {
...
+        return computePublicKey(publicKey_digest, true).slice(2);
+      }
+
+      // In the event that the wallet isn't hashing message as per spec, attempt recovery without digest
+      const publicKey_hash = recoverPublicKey(hash, sig);
+      if (address === computeAddress(publicKey_hash)) {
```
adamstaveley, line 325:
```suggestion
      if (getAddress(address) === getAddress(computeAddress(publicKey_hash))) {
```

src/iam/iam-base.ts
```diff
@@ -306,17 +306,25 @@ export class IAMBase {
     }
   }
 
-  private async getPublicKey() {
+  private async getPublicKey(): Promise<string> {
     if (this._signer) {
       const address = await this._signer.getAddress();
       const hash = keccak256(address);
+      const sig = await this._signer.signMessage(arrayify(hash));
```
adamstaveley, line 313: If we are not prefixing here, is there any point in hashing the address? Couldn't we just send `arrayify(address)` to the wallet? The user might then have a better idea of what they're signing. But maybe I'm overlooking something with this.

adamstaveley: Additionally, what do you think of `eth_signTypedData`? I think it would be a good solution to the problems we face with `eth_sign`, though I'm not sure how widespread it's supported (and indeed we would need to support it in KM).

src/iam/iam-base.ts
jrhender: Thanks, I added `getAddress`

src/iam/iam-base.ts
jrhender: Haha, I thought of not hashing the address as well, but I was scared to change it in case there was a good reason. But I also can't think of a good reason and it would be nice if the user could recognize what they are hashing. @dwojno or @JGiter do you know why the address was hashed?

jrhender: Re `eth_signTypedData`, I think we only have access to the ethers (v4) Signer api, which doesn't have _signTypedData https://docs.ethers.io/v4/api-wallet.html#signer . Ethers v5 does (https://docs.ethers.io/v5/api/signer/#Signer-signTypedData) but it seems experimental so maybe a bit earlier to rely on at this point and as you say, it might not be supported widely by wallets.

JGiter: It looks like there's really no need to hash the address

PR #102, opened 2021-02-03
feat(nats): add flag to indicate that nats is not connected

dwojno:


kim-energyweb (approved):
LGTM

src/iam/iam-base.ts
```diff
@@ -337,12 +337,25 @@ export class IAMBase {
...
-      let protocol = "ws";
-      if (this._natsServerUrl.indexOf("https://") === 0) {
-        protocol = "wss";
+      try {
+        let protocol = "ws";
+        if (this._natsServerUrl.indexOf("https://") === 0) {
+          protocol = "wss";
+        }
+
+        const connection = await Promise.race<NatsConnection | undefined>([
```
jrhender, line 346: It looks to me like this is adding kind-of a back-up timeout to the connect. Is that right? As it looks like the `connect` constructor already has a timeout. Why the need for the duplication? Is it that the `connect` timeout is unreliable?
Let me know if I'm missing something or it's a common pattern that I just don't know 😅

dwojno: xD yeah, this NATS.ws is throwing error on timeout that I'm not able to catch ;) I've tried all things I can think off, try/catch in all places, .catch(), nothing works, so I've added a race condition timeout to make it work :)

src/iam/iam-base.ts
```diff
@@ -337,12 +337,25 @@ export class IAMBase {
...
+          connect({
+            servers: `${protocol}://${this._natsServerUrl}`,
+            timeout: 2000
```
jrhender, line 349: Minor point, but I would consider assigning `2000` to a variable so that both timeouts can be sure to use the same thing

jrhender (approved):
Looks good

src/iam/iam-base.ts
dwojno: 👍

PR #107, opened 2021-02-04
feat(metamask): return more info from isMetamaskExtensionPresent

dwojno:
BREAKING CHANGE: now isMetamaskExtensionPresent method
returns { isMetamaskPresent: boolean, chainId?: number }
instead of boolean


jrhender (approved):
LGTM

JGiter:
Since this PR contains breaking change it can be reflected in new major version which will be published if commit message has `BREAKING CHANGE`

PR #108, opened 2021-02-04
feat(claim): implement cancel claim request

dwojno:


src/iam.ts
```diff
@@ -1204,6 +1204,13 @@ export class IAM extends IAMBase {
     this._natsConnection.publish(`${requesterDID}.${NATS_EXCHANGE_TOPIC}`, dataToSend);
   }
 
+  async cancelClaimRequest({ id }: { id: string }) {
```
jrhender, line 1207: Personally, I think having the method name as `deleteClaim` is better. "CancelClaim" to me implies that it is just the status of the request that is changing and the request still exists but in reality, the claim is being deleted so better just to say that.

dwojno: 👍

JGiter (approved):

PR #113, opened 2021-02-08
feat(kms integration): reuse existing tab

jrhender:
Note that this reloads the existing KMS tab (if it exists) and could interrupt the user. Do you think this is a good UX?

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -271,7 +271,7 @@ export class IAMBase {
         const encoded = encodeURIComponent(wcUri);
         const hasQueryString = this._ewKeyManagerUrl.includes("?");
         const url = `${this._ewKeyManagerUrl}${hasQueryString ? "&" : "?"}uri=${encoded}`;
-        window.open(url, "_blank");
+        window.open(url, "ew_key_manager");
       });
     }
 


src/iam/iam-base.ts
```diff
@@ -271,7 +271,7 @@ export class IAMBase {
         const encoded = encodeURIComponent(wcUri);
         const hasQueryString = this._ewKeyManagerUrl.includes("?");
         const url = `${this._ewKeyManagerUrl}${hasQueryString ? "&" : "?"}uri=${encoded}`;
-        window.open(url, "_blank");
+        window.open(url, "ew_key_manager");
       });
```
JGiter, line 275: is "ew_key_manager" a name of the window? As I understand to open in the same tab the '_parent' name is used

jrhender: Yes, I think that parameter is just for a descriptive window name. Taking a look at the MDN docs (https://developer.mozilla.org/en-US/docs/Web/API/Window/open) for `window.open`, they've got an example (https://developer.mozilla.org/en-US/docs/Web/API/Window/open#examples), where they just use `"CNN_WindowName"`
image

They also mention that `_blank` is a special value:
>To open a new window on every call of window.open(), use the special value _blank for windowName.

I don't think the MDN docs mention `_parent` but after testing, it seems to cause `window.open` to use the same tab which is consistent with https://stackoverflow.com/a/27271624/6674421 (and which I don't think is what we want)

JGiter (approved):

PR #114, opened 2021-02-09
feat(login): store session information on iam side

dwojno:
This PR contains changes required to move controle of wallet's session to iam-client-lib:
- storing information about current session(publicKey, walletProvider)
- authentication using cookie
- handling networkChange, accountChange and disconnect events
- add static methods to get info about current session

PR needs to merge after [iam-cache-server PR](https://github.com/energywebfoundation/iam-cache-server/pull/65) .

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -27,7 +27,7 @@ import {
 import { hashes, IProofData, ISaltedFields } from "@ew-did-registry/claims";
 import { namehash } from "./utils/ENS_hash";
 import { v4 as uuid } from "uuid";
-import { IAMBase, emptyAddress } from "./iam/iam-base";
+import { IAMBase, emptyAddress, WALLET_PROVIDER, PUBLIC_KEY } from "./iam/iam-base";
 import {
   CacheClientNotProvidedError,
   ChangeOwnershipNotPossibleError,
@@ -99,6 +99,21 @@ export class IAM extends IAMBase {
 
     return { isMetamaskPresent: !!provider, chainId };
   }
+
+  static isSessionActive() {
+    if (window && window.sessionStorage) {
+      return Boolean(sessionStorage.getItem(PUBLIC_KEY));
+    }
+    return false;
+  }
+
+  static getProviderType() {
+    if (window && window.sessionStorage) {
+      const providerType = sessionStorage.getItem(WALLET_PROVIDER) as WalletProvider | null;
+      return providerType;
+    }
+    return null;
+  }
   // GETTERS
 
   /**
@@ -132,7 +147,7 @@ export class IAM extends IAMBase {
    * @returns did string, status of connection and info if the user closed the wallet selection modal
    */
   async initializeConnection({
-    walletProvider,
+    walletProvider = this._providerType,
     reinitializeMetamask
   }: { walletProvider?: WalletProvider; reinitializeMetamask?: boolean } = {}): Promise<
     InitializeData
@@ -180,9 +195,13 @@ export class IAM extends IAMBase {
   async closeConnection() {
     if (this._walletConnectProvider) {
       await this._walletConnectProvider.close();
-      this._did = undefined;
-      this._address = undefined;
-      this._signer = undefined;
+    }
+    this.clearSession();
+    this._did = undefined;
+    this._address = undefined;
+    this._signer = undefined;
+    if (this._runningInBrowser) {
+      location.reload();
     }
   }
 
@@ -856,7 +875,6 @@ export class IAM extends IAMBase {
     if (returnSteps) {
       return steps;
     }
-
     await this.send({ calls: steps.map(s => s.tx), from });
     return [];
   }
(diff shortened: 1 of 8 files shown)


src/iam/iam-base.ts
```diff
@@ -211,27 +227,21 @@ export class IAMBase {
...
-            ]
+    if (walletProvider === WalletProvider.MetaMask) {
+      const metamaskProvider: any = await detectMetamask({
+        mustBeMetaMask: true
+      });
+      if (!metamaskProvider) {
+        throw new Error(ERROR_MESSAGES.METAMASK_EXTENSION_NOT_AVAILABLE);
+      }
+      const requestObject = {
+        method: initializeMetamask ? "wallet_requestPermissions" : "eth_accounts",
```
jrhender, line 238: 👍

src/iam/iam-base.ts
```diff
@@ -245,38 +255,97 @@ export class IAMBase {
...
+      this._providerType = walletProvider;
+      return;
     }
+    throw new Error(ERROR_MESSAGES.WALLET_TYPE_NOT_PROVIDED);
+  }
 
-    await this._walletConnectProvider.enable();
-    this._signer = new providers.Web3Provider(this._walletConnectProvider).getSigner();
+  private storeSession() {
+    if (this._runningInBrowser && window && this._didSigner) {
```
jrhender, line 319: Should this check for `window` should be included in the `runningInBrowser` check? Or is it only relevant in this case and in `clearSession`?

src/iam/iam-base.ts
```diff
@@ -245,38 +255,97 @@ export class IAMBase {
...
+
+      await this._walletConnectProvider.enable();
+
+      this._walletConnectProvider.wc.on("session_update", () => {
+        this.closeConnection();
+      });
+
+      this._walletConnectProvider.wc.on("disconnect", () => {
+        this.closeConnection();
+      });
```
jrhender, line 309: Just curious, are these eventListeners set after `walletConnectProvider.enable` for a reason? In a situation where disconnect is emitted immediately after enable, would we want to be listening for that event?

src/iam.ts
```diff
@@ -98,6 +98,21 @@ export class IAM extends IAMBase {
...
+    return false;
+  }
+
+  static getProviderType() {
+    if (window && window.sessionStorage) {
+      const providerType = sessionStorage.getItem(WALLET_PROVIDER) as WalletProvider | null;
+      return providerType;
+    }
+    return null;
+  }
```
jrhender, line 115: Just curious, why are are these functions static? (maybe there is a good reason that I just can't see 😄 ) I'm just thinking that if they were instance members, then they could use the `_providerType` and `_publicKey` properties which would mean that the use of sessionStorage would be a little more encapsulated?

src/iam/iam-base.ts
```diff
@@ -245,38 +255,97 @@ export class IAMBase {
...
+  async closeConnection() {
+    if (this._walletConnectProvider) {
+      await this._walletConnectProvider.close();
+    }
+    this.clearSession();
+    this._did = undefined;
+    this._address = undefined;
+    this._signer = undefined;
+    if (this._runningInBrowser) {
+      location.reload();
```
jrhender, line 347: I'm a little hesitant to be having the client-lib reloading apps which are depending on it. Not sure if it's in the scope of this story, but maybe it would be better to have the client-lib be an eventEmitter and emit a "closed" event and then apps can reload if they want? I think we might have discussed doing something like this at some point recently.

src/iam.ts
```diff
@@ -98,6 +98,21 @@ export class IAM extends IAMBase {
 
     return { isMetamaskPresent: !!provider, chainId };
   }
+
+  static isSessionActive() {
+    if (window && window.sessionStorage) {
+      return Boolean(sessionStorage.getItem(PUBLIC_KEY));
```
jrhender, line 104: Just curious, is there is a reason in particular why having the publicKey set is more indicative of the session state than the provider? I think this definitely looks good and works as is, just a trying to understand a little better

jrhender (commented):
Looks really good! I think having iam-client-lib do the session management will be really nice. Left a few comments/questions.

JGiter (approved):

src/iam/iam-base.ts
dwojno: Yeah, good point, we don't need to check if `window` is there cause the `runningInBrowser` is checking for env, any every browser has a `window`.

src/iam/iam-base.ts
dwojno: After second though the listeners don't need to be set after the enabling the `walletConnectProvider` , in this spring we could add a `on` method to be able to add listeners for certain events.

src/iam.ts
dwojno: Yeah good point, I'll do that 👍

src/iam/iam-base.ts
dwojno: Good point 👍  I'll remove this for now, and afterwards we can add this `on` method

src/iam.ts
dwojno: Yeah I think you're right to be able to return true here we need to have both information like publicKey and provider type, I'll change that 👍

jrhender (approved):

PR #118, opened 2021-02-15
fix(exchange): multiple connections to NATS websocket

dwojno:
Our nats exchange server is behind cloudflare, and cloudflare is closing websocket connection after 100 seconds of idleness.
This PR basicly changes the ping interval to 90 seconds to omit closing connection by cloudflare.

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -429,7 +429,8 @@ export class IAMBase {
         const connection = await Promise.race<NatsConnection | undefined>([
           connect({
             servers: `${protocol}://${this._natsServerUrl}`,
-            timeout
+            timeout,
+            pingInterval: 50 * 1000
           }),
           new Promise<undefined>(resolve => setTimeout(resolve, timeout))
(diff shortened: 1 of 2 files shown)


src/iam/iam-base.ts
```diff
@@ -429,7 +429,8 @@ export class IAMBase {
         const connection = await Promise.race<NatsConnection | undefined>([
           connect({
             servers: `${protocol}://${this._natsServerUrl}`,
-            timeout
+            timeout,
+            pingInterval: 50 * 1000
```
jrhender, line 433: Probably not a big deal, but you mentioned 90 seconds in your description of the change and this seems to be 50 seconds. Is that okay/intentional?

dwojno: Yeah, sorry it was 90 seconds, but I've noticed that this ping interval is not always 90 seconds exactly, it may be longer, so I've changed it to more secure value :)

jrhender (approved):

PR #123, opened 2021-02-19
include isOwned to org hieararchy

JGiter:
To enable ownership checking in extended version of Iam return type of `getOrgHieararchy` is appended with  `isOwned` property

### DIFF
--- a/src/GnosisIam.ts
+++ b/src/GnosisIam.ts
@@ -141,7 +141,7 @@ export class GnosisIam extends IAM {
       });
   }
 
-  async getOrgHierarchy({ namespace }: { namespace: string }) {
+  async getOrgHierarchy({ namespace }: { namespace: string }): Promise<IOrganization> {
     if (!this._cacheClient) {
       throw new CacheClientNotProvidedError();
     }
--- a/src/GnosisIam.ts
+++ b/src/GnosisIam.ts
@@ -147,7 +147,7 @@ export class GnosisIam extends IAM {
     }
     const org = await this._cacheClient.getOrgHierarchy({ namespace });
     [org, ...org.subOrgs || [], ...org.apps || [], ...org.roles || []]
-      .forEach((domain) => domain.isOwned = [this._address, this.safeAddress].includes(domain.owner));
+      .forEach((domain) => domain.isOwnedByCurrentUser = [this._address, this.safeAddress].includes(domain.owner));
     return org;
   }
 }
--- a/src/GnosisIam.ts
+++ b/src/GnosisIam.ts
@@ -140,4 +140,14 @@ export class GnosisIam extends IAM {
         notOwnedNamespaces: new Array<string>()
       });
   }
+
+  async getOrgHierarchy({ namespace }: { namespace: string }) {
+    if (!this._cacheClient) {
+      throw new CacheClientNotProvidedError();
+    }
+    const org = await this._cacheClient.getOrgHierarchy({ namespace });
+    [org, ...org.subOrgs || [], ...org.apps || [], ...org.roles || []]
+      .forEach((domain) => domain.isOwned = [this._address, this.safeAddress].includes(domain.owner));
+    return org;
+  }
 }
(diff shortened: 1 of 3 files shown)


src/iam.ts
```diff
@@ -1084,11 +1084,29 @@ export class IAM extends IAMBase {
...
+  async getOrgHierarchy({ namespace }: { namespace: string }) {
     if (!this._cacheClient) {
       throw new CacheClientNotProvidedError();
     }
-    return this._cacheClient.getOrgHierarchy({ namespace });
+    const org = await this._cacheClient.getOrgHierarchy({ namespace });
+    const { subOrgs, apps, roles } = org;
+    if (subOrgs?.length && subOrgs.length > 0) {
+      org.subOrgs = await Promise.all(subOrgs.map(
+        async (o) => ({ ...o, isOwned: await this.isOwner({ domain: o.namespace }) })
```
kim-energyweb, line 1095: Is there a way to prevent calling extra backend service? **iam-client-lib** is still part of frontend.

kim-energyweb (changes requested):

src/iam.ts
JGiter: Right. I just wanted to reuse already overridden `isOwner`, but to avoid hops also need to override `getOrgHieararchy`

src/cacheServerClient/cacheServerClient.types.ts
```diff
@@ -55,6 +56,7 @@ export interface IOrganization {
   apps?: IApp[];
   roles?: IRole[];
   subOrgs?: IOrganization[];
+  isOwned?: boolean;
```
jrhender, line 59: Why is it only nullable in the case of Organization (and not nullable for Role and App)? Is it because it doesn't have meaning in the context of top-level org? If so, maybe add a comment to explain the reason

src/GnosisIam.ts
```diff
@@ -140,4 +140,14 @@ export class GnosisIam extends IAM {
         notOwnedNamespaces: new Array<string>()
       });
   }
+
+  async getOrgHierarchy({ namespace }: { namespace: string }) {
```
jrhender, line 144: Wouldn't hurt to have the return type on this method as well

src/iam.ts
```diff
@@ -1084,11 +1085,14 @@ export class IAM extends IAMBase {
...
    */
-  getOrgHierarchy({ namespace }: { namespace: string }) {
+  async getOrgHierarchy({ namespace }: { namespace: string }): Promise<IOrganization> {
     if (!this._cacheClient) {
       throw new CacheClientNotProvidedError();
     }
-    return this._cacheClient.getOrgHierarchy({ namespace });
+    const org = await this._cacheClient.getOrgHierarchy({ namespace });
+    [org, ...org.subOrgs || [], ...org.apps || [], ...org.roles || []]
+      .forEach((domain) => domain.isOwned = domain.owner === this.address);
```
jrhender, line 1094: I find the `isOwned` name to be a little confusing. At first glance, I think it implies that domain is owned at all. Like I think the obvious guess when first seeing the property name is that `isOwned == false` means that the domain isn't owned at all. But it seems what it really means it that isn't owned by the current user. So maybe `isOwnedByCurrentUser` is better, though it is starting to get a little wordy 🤔

jrhender (approved):
Looks good but added some suggestions on how it could be made more clear

src/cacheServerClient/cacheServerClient.types.ts
JGiter: Thanks for noticing. Will fix

src/iam.ts
JGiter: I agree. This is a more appropriate name. @kim-energyweb , what do you think of this?

kim-energyweb: @JGiter I agree. This is a more concise attribute name.

kim-energyweb (changes requested):
Aside from the attribute name suggested by John, it looks better now.

kim-energyweb (approved):
LGTM

PR #124, opened 2021-02-22
feature/make switchboard chain agnostic

JGiter:
Changes:
- [x] `chainId`, `ensRegistryAddress`, `ensResolverAddress`, `cacheClient` are excluded from IamBase constructor parameters because all of them can be resolved from provider or `rpcUrl`
- [x] chain configurations are set separately from `IAMBase` definition
- [x] test environment changed from browser to nodejs, because testing in browser would require using MetaMask or WalletConnect

Propositions:
- Some of the Iam properties types like `_ensRegistry` includes `undefined` while this is not essentially their type. The only reason for this - is typescript requirement to define all non-null properties in constructor. This requirement can be lifted by setting tsconfig `strictPropertyInitialization` property to `false`


src/iam/iam-base.ts
```diff
@@ -213,10 +213,8 @@ export class IAMBase {
         signer: this._signer,
         publicKey: this._publicKey
       });
-      this._publicKey = this._didSigner.publicKey;
```
dwojno, line 216: I think this should stay, that was added on purpose, cause after logging in the publicKey is stored only on session storage but not on the this._publicKey property, so when calling this._isSessionActive() it returns false.

src/iam/iam-base.ts
```diff
@@ -100,11 +105,11 @@ export class IAMBase {
   protected _safeAddress: string | undefined;
   protected _didSigner: IdentityOwner | undefined;
   protected _transactionOverrides: TransactionOverrides = {};
-  protected _providerType: WalletProvider | undefined;
-  protected _publicKey: string | undefined;
+  protected _providerType: WalletProvider;
+  protected _publicKey: string;
```
dwojno, line 109: I think both of them should by `type | null` cause when something is not stored in sessionStorage the browser API will return null

JGiter: Ok

src/iam/iam-base.ts
JGiter: Right. I didn't noticed that pub key used to check session status. Will put it back

src/iam/iam-base.ts
```diff
@@ -279,11 +279,12 @@ export class IAMBase {
...
 
       this._providerType = walletProvider;
+      this._provider = this._signer.provider as providers.Web3Provider;
+      console.log('metamask chain id:', (await this._provider.getNetwork()).chainId);
       return;
     }
     if (
-      walletProvider === WalletProvider.EwKeyManager ||
-      walletProvider === WalletProvider.WalletConnect
+      walletProvider && [WalletProvider.EwKeyManager, WalletProvider.WalletConnect].includes(walletProvider)
```
jrhender, line 287: 👍

src/iam/iam-base.ts
```diff
@@ -673,4 +677,31 @@ export class IAMBase {
...
+    };
+    this._ensRegistryAddress = ensRegistryAddress;
+    this._ensResolverAddress = ensResolverAddress;
+    this._ensRegistry = EnsRegistryFactory.connect(ensRegistryAddress, this._provider);
+    this._ensResolver = PublicResolverFactory.connect(ensResolverAddress, this._provider);
+    cacheServerUrl && (this._cacheClient = new CacheServerClient({ url: cacheServerUrl }));
+  }
+
+  /**
+   * @description - should be invoked before iniitialization
```
jrhender, line 699: small typo "initialization"

src/iam/iam-base.ts
```diff
@@ -673,4 +677,31 @@ export class IAMBase {
...
+   * 
+   * @param chainId 
+   * @param config 
+   */
+  static setChainConfig(chainId: number, config: ChainConfig) {
```
jrhender, line 704: Cool approach 🧠 . Just curious, why this approach rather than passing into the constructor? I think it could be nice, it just deviates a bit from the "config in the constructor" approach that we have so far?

jrhender, line 704: Not a big deal, but I think it would be nice if could selectively override a property for a given config, instead of needing to specific the whole config.  Maybe consider accepting a `Partial<ChainConfig>`.

jrhender, line 704: Just a suggestion, but I think would be nice to add some info on how this works to the documentation.

src/iam/iam-base.ts
```diff
@@ -673,4 +677,31 @@ export class IAMBase {
...
+    this._registrySetting = {
+      address: didContractAddress,
+      abi: abi1056,
+      method: Methods.Erc1056
+    };
+    this._ensRegistryAddress = ensRegistryAddress;
+    this._ensResolverAddress = ensResolverAddress;
+    this._ensRegistry = EnsRegistryFactory.connect(ensRegistryAddress, this._provider);
+    this._ensResolver = PublicResolverFactory.connect(ensResolverAddress, this._provider);
+    cacheServerUrl && (this._cacheClient = new CacheServerClient({ url: cacheServerUrl }));
```
jrhender, line 695: I think this is a bit of a problem, as `CacherServerClient` has another initialization property (`cacheServerSupportAuth`) and could have others in the future. Maybe it is better just to continue to pass in the `CacheServerClient` in the constructor for now?

src/iam/iam-base.ts
```diff
@@ -67,6 +70,14 @@ export type Transaction = {
   from: string;
 };
 
+export type ChainConfig = {
+  rpcUrl: string,
```
jrhender, line 74: If we're requiring the `rpcUrl` in `ChainConfig` (which I think we need for WalletConnect, right?), could we remove it from `ConnectionOptions`? I think it is confusing that it is in both places

src/iam/iam-base.ts
```diff
@@ -67,6 +70,14 @@ export type Transaction = {
   from: string;
 };
 
+export type ChainConfig = {
```
jrhender, line 73: Maybe consider changing this to an interface so that it shows in TypeDoc documentation. I don't think types get picked up, but interfaces do?

jrhender (commented):
Sorry, left quite a few comments 😅  Overall, looks really good, I think it adds some nice encapsulation around the chain-specific config

src/iam/iam-base.ts
```diff
@@ -41,7 +43,8 @@ export enum MessagingMethod {
 }
 
 export type ConnectionOptions = {
-  rpcUrl: string;
+  // only required in node env
```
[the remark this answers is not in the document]
JGiter: Useful thing

src/iam/iam-base.ts
JGiter: Unfortunately we need both of them, because the one which is defined in `ConnectionOptions` along with `privateKey` is a way to connect to blockchain instead of providing wallet. `rpcUrl` defined in `ChainConfig` is not input value, but part of the mapping used by walled to resolve url from chain id

src/iam/iam-base.ts
JGiter: Since there are just a few possible chain configurations its make sense to set them as defaults. `setChainConfig` is supposed to be used to override or add missing configuration. Perhaps some documentation will be useful

JGiter: Good idea

tsconfig.json
```diff
@@ -28,7 +28,7 @@
     "strict": true,
     "strictFunctionTypes": true,
     "strictNullChecks": true,
-    "strictPropertyInitialization": true,
+    "strictPropertyInitialization": false,
```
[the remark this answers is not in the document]
JGiter: @jrhender This options forces to append `undefined` to types of properties which can't be initialized in constructor (like `_ensRegisty`) and later always check them before using. It seems to me that these checking are not very useful, because they are just suppresses errors which indicates incorrect `IAM` initialization. Its a debatable question though. What do you think about it @dwojno ?

src/iam/iam-base.ts
jrhender: Hmm, I think I see what you're saying. You're saying that by constructing a default cache-server like this, if for example, the user changed their network in metamask, we could connect to the correct cache-server for them.

How about this design? 
- Have an interface of `CacheServerClientOptions` that is used for `CacheServerClient` initialization
- Have `ChainConfig` extend `CacheServerClientOptions`
- In this way, we can have a default for each chain, but also apps can override their cache-server config if they want

src/iam/extras.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+};
+
+/**
+ * Used to override existing configuration or add a missing one
+ * Configuration must be set before constructing `IAM`
+ */
+export const setChainConfig = function (chainId: number, config: Partial<ChainConfig>) {
+  chainConfigs[chainId] = { ...chainConfigs[chainId], ...config };
+};
```
jrhender, line 25: I feel kindof bad as you already changed it and this is just a matter of opinion 😅 . But I would rename this file to `ChainConfig.ts` (so that the name is more descriptive) and just move the ChainConfig interface into this file.

tsconfig.json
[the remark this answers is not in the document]
jrhender: Hmm, okay. In my opinion, it's good to be forced to append `undefined` if there is a possibility that it could be undefined. But if you guys both think it's unnecessary, then we can turn it off.

src/iam/extras.ts
JGiter: Ok

src/iam/iam-base.ts
JGiter: Agree. Only suggestion - is to move cache client server options from chain config to allow partial update of both of them

src/iam/chainConfig.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const setChainConfig = function (chainId: number, config: Partial<ChainConfig>) {
+  chainConfigs[chainId] = { ...chainConfigs[chainId], ...config };
+};
+
+/**
+ * Used to override existing cache server configuration or add a missing one
+ * Configuration must be set before constructing `IAM`
+ */
+export const setCacheClientOptions = function (chainId: number, options: Partial<CacheServerClientOptions>) {
```
jrhender, line 43: 👍

src/iam/iam-base.ts
```diff
@@ -673,4 +658,23 @@ export class IAMBase {
...
+      abi: abi1056,
+      method: Methods.Erc1056
+    };
+    this._ensRegistryAddress = ensRegistryAddress;
+    this._ensResolverAddress = ensResolverAddress;
+    this._ensRegistry = EnsRegistryFactory.connect(ensRegistryAddress, this._provider);
+    this._ensResolver = PublicResolverFactory.connect(ensResolverAddress, this._provider);
+
+    const cacheOptions = cacheServerClientOptions[chainId];
+    cacheOptions.url && (this._cacheClient = new CacheServerClient(cacheOptions));
```
jrhender, line 678: 👍

jrhender (approved):

PR #126, opened 2021-02-23
feat(cacheClient): add option to disable auth

dwojno:


src/cacheServerClient/cacheServerClient.ts
```diff
@@ -10,15 +10,23 @@ export class CacheServerClient implements ICacheServerClient {
...
   private failedRequests: Array<() => void> = [];
+  private authEnabled: boolean;
 
-  constructor({ url }: { url: string }) {
+  constructor({
+    url,
+    cacheServerSupportsAuth = true
+  }: {
+    url: string;
+    cacheServerSupportsAuth?: boolean;
```
jrhender, line 20: @dwojno How does the consumer (e.g. Switchboard app) of iam-client-lib set this `cacheServerSupportsAuth` param?

dwojno: Like this:
```
        const cacheClient = new CacheServerClient({ url: 'http://cache-server.com', cacheServerSupportsAuth: false })

        this._iam = new IAM({
          rpcUrl: 'https://volta-rpc.energyweb.org/',
          chainId: 73799,
          cacheClient // optional
        });
```

jrhender: Ah right, I forgot they pass in the cache server client.

jrhender (approved):
👍

PR #130, opened 2021-03-01
feat(auth): add auth support in nodejs env

dwojno:


src/cacheServerClient/cacheServerClient.ts
```diff
@@ -27,13 +31,23 @@ export class CacheServerClient implements ICacheServerClient {
       return response;
     }, this.handleUnauthorized);
     this.authEnabled = cacheServerSupportsAuth;
+    this.isBrowser = isBrowser();
   }
 
-  handleSuccessfulReLogin() {
-    this.failedRequests = this.failedRequests.filter(callback => callback());
+  async handleSuccessfulReLogin() {
```
jrhender, line 37: I would change this method name to `handleReLogin` as it isn't successful yet

jrhender (approved):
Looks good, thanks for fixing

PR #131, opened 2021-03-01
feat(did): get did doc from chain when cache server fails

dwojno:


src/iam.ts
```diff
@@ -215,11 +215,29 @@ export class IAM extends IAMBase {
...
-        service: didDoc.service as (IServiceEndpoint & ClaimData)[]
-      };
+      try {
+        const didDoc = await this._cacheClient.getDidDocument({ did, includeClaims });
+        return {
+          ...didDoc,
+          service: didDoc.service as (IServiceEndpoint & ClaimData)[]
+        };
+      } catch (err) {
+        if (err?.response?.status === 404) {
```
jrhender, line 225: I don't think we could catch only on 404, I think we should catch any error and just fallback to the blockchain. Does that seem reasonable?

src/iam.ts
```diff
@@ -215,11 +215,29 @@ export class IAM extends IAMBase {
...
+              ...document,
+              service: includeClaims
+                ? ((await this.downloadClaims({
+                    services:
+                      document.service && document.service.length > 0 ? document.service : []
+                  })) as (IServiceEndpoint & ClaimData)[])
+                : []
+            };
+          }
+        }
```
jrhender, line 238: Instead of duplicating the code here, I would just fall through to the code below, I think it is the same.

jrhender (changes requested):
A couple suggestions to address

src/iam.ts
dwojno: Ok, let's fallback to blockchain on every error except 401

src/iam.ts
dwojno: 👍

src/iam.ts
jrhender: I think there's a case to be made for getting from chain even if you get a 401 from the cache-server. To me, it doesn't really sense that iam-client-lib would say "oh, you're not authorized for the cache-server, so I'm not going to let you fallback to the blockchain directly". What do you think?

dwojno: Yeah, your right 👍  there is no need to be authorized to get something from chain, I'll change that

manihagh (approved):

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "noLib": false,
-    "noUnusedLocals": false,
+    "noUnusedLocals": true,
```
jrhender, line 21: 👍

jrhender (approved):
Nice work!

PR #132, opened 2021-03-02
feature/change resolver script

JGiter:
Added script to change resolver in all nodes under some domain. The script is intended to update the logic of the resolver while preserving the existing nodes. Script can used both programmatically and as cli. To execute script run `npm run change_resolver`. For example of use `tests/utilsTests.ts`. Along with nodes also migrated node names and text. Test timeout have been doubled.


src/utils/change_resolver.ts
```diff
@@ -0,0 +1,58 @@
...
+      owners[node] = owner;
+
+      await ensRegistry.setOwner(namehash(node), rootOwner);
+      await ensRegistry.setResolver(namehash(node), newResolverAddr);
+    }
+  };
+
+  const restoreOwners = async () => {
+    for (const node in owners) {
+      await ensRegistry.setOwner(namehash(node), owners[node]);
```
jrhender, line 51: I'm surprised this works as the rootOwner might not own intermediary nodes anymore. See this test I made: https://github.com/energywebfoundation/iam-contracts/blob/af60a5ad99cc763a3ccf14b668ed675e51d8227f/test/RoleDefinitionResolver.Test.ts#L104
Maybe I'm missing something though 🤔

jrhender (commented):
Thanks for working on this! Will come super in handy!

src/utils/change_resolver.ts
JGiter: `changeDomainResolver` traverses all nodes under parent and changes their owner to on `rootOwner` in order to be able to change resolver

jrhender: I could be wrong but I think this test should pass, but it fails. Please try it out.

```
export const utilsTests = () => {
  const newResolverAddr = emptyAddress;

  test('resolver can be changed', async () => {
    const rootNode = `${root}`; // Changed to just root...

    const params = {
      rootNode,
      privateKey: rootOwner.privateKey,
      registryAddr: ensRegistry.address,
      resolverAddr: ensResolver.address,
      rpcUrl,
      newResolverAddr
    };

    // Print out current owners (all same address)
    const actualRootOwner = await ensRegistry.owner(namehash(root));
    console.info("actual root owner", actualRootOwner);
    const org1owner = await ensRegistry.owner(namehash(rootNode));
    console.info("org1 owner", org1owner);
    const domains_1 = await getSubdomains({ domain: rootNode, ensResolver, ensRegistry, mode: 'ALL' });
    await Promise.all(domains_1.map(async domain => {
      const owner = await ensRegistry.owner(namehash(domain));
      console.info(`${domain} owner`, owner);
    }));

    // Set to new address (so orginal address owns the root, doesn't own the org or the sub org)
    await ensRegistry.setOwner(namehash('org1-1.org1.root'), '0xE45Ad1e7522288588dA6829A9ea6A09e92FCDe14');
    await ensRegistry.setOwner(namehash('org1.root'), '0xE45Ad1e7522288588dA6829A9ea6A09e92FCDe14');

    await changeResolver(params);

    const domains = await getSubdomains({ domain: rootNode, ensResolver, ensRegistry, mode: 'ALL' });
    const resolvers = await Promise.all(domains.map(async (domain) => await ensRegistry.resolver(namehash(domain))));

    expect(
      resolvers
        .every((addr) => addr === newResolverAddr))
      .toBe(true);
  });
};
```

jrhender:
@JGiter I was wondering if data migration should be a part of this script was well? Maybe:
step 1: migrate data to the new resolver
step 2: point registry to new resolver
What do you think?

src/utils/change_resolver.ts
JGiter: Thanks. You have found serious bug. I should have changed the owner using `setSubnodeOwner` and not `setOwner`

JGiter:
> @JGiter I was wondering if data migration should be a part of this script as well? Maybe:
> step 1: migrate data to the new resolver
> step 2: point registry to new resolver
> What do you think? We can add in a separate story perhaps, just wondering what you think

@jrhender I agree, its useful addition. I will add it

test/utils.testSuite.ts
```diff
@@ -0,0 +1,53 @@
...
+    const domains = await getSubdomains({ domain: params.rootNode, ensResolver, ensRegistry, mode: 'ALL' });
+    const resolvers = await Promise.all(domains.map(async (domain) => await ensRegistry.resolver(namehash(domain))));
+
+    expect(
+      resolvers
+        .every((addr) => addr === newResolverAddr))
+      .toBe(true);
+  });
+
+  test('root resolver can be changed', async () => {
```
jrhender, line 37: For this test, I think it would be good to also add an assertion that the ownership at the end is the same as it was at the beginning. In other words, expect that at the end `org1-1.org1.root` and `org1.root` are owned by `0xE45Ad1e7522288588dA6829A9ea6A09e92FCDe14` and the other domains are owned by the other account. What do you think?

JGiter: Agree

dwojno (approved):

jrhender (approved):
Looks good

PR #133, opened 2021-03-03
feat(enrolment): add enrolment precondition verification

dwojno:


src/cacheServerClient/cacheServerClient.types.ts
```diff
@@ -20,6 +20,7 @@ export interface IRoleDefinition {
     did?: string[];
     roleName?: string;
   };
+  enrolmentPreconditions: { type: string; value: string[] }[];
```
jrhender, line 23: Instead of `type: string`, could `type` be an enum?

manihagh (approved):

src/iam/chainConfig.ts
```diff
@@ -24,22 +24,26 @@ export const chainConfigs: Record<number, ChainConfig> = {
 
 export const cacheServerClientOptions: Record<number, CacheServerClientOptions> = {
   [VOLTA_CHAIN_ID]: {
-    url: "https://identitycache-dev.energyweb.org/"
+    url: "https://volta-identitycache.energyweb.org/",
+    cacheServerSupportsAuth: true
```
jrhender, line 28: 😎

src/iam/iam-base.ts
```diff
@@ -131,7 +131,8 @@ export class IAMBase {
     natsServerUrl,
     bridgeUrl = "https://walletconnect.energyweb.org",
     privateKey,
-    ewKeyManagerUrl = "https://km.aws.energyweb.org/connect/new"
+    ewKeyManagerUrl = "https://km.aws.energyweb.org/connect/new",
+    cacheClient
```
jrhender, line 135: I'm a little confused on whether or not we should we adding back the `cacheClient` in the constructor. I think @JGiter and I discussed this and we settled on using the `setCacheClientOptions` function that is in `chainConfig.ts`https://github.com/energywebfoundation/iam-client-lib/pull/124#discussion_r582541327 . So I think if we add it back, then we will have several ways of setting the `cacheClient` config. Maybe this is okay but maybe it is better to have one way. @JGiter Any thoughts?

jrhender (changes requested):
The enrolment precondition verification part looks good. I think we should just figure out our "API" for configuring the cacheClient so that there isn't any unnecessary duplication. Does that seem reasonable or am I making a mountain out of molehill here?

src/iam/iam-base.ts
dwojno: ahhh ok 👍  I'll change it back, but still we need to update the readme ;)

dwojno: @jrhender @JGiter but you need to remember that in this implementation the iam-client-lib user is not able to pass his own cacheServerClient implementation, they need to use our one.

jrhender: @dwojno That's a great point, I hadn't thought of that tbh 😅 . How about having an optional parameter that is a factory function: `cacheClientFactory?: (CacheServerClientOptions) => ICacheServerClient`. In this way, we still rely on the `setCacheClientOptions` to set the cache-server config, but also give the option to override the implementation. What do you think?

kim-energyweb (approved):

jrhender (approved):

PR #134, opened 2021-03-05
feat(messaging): make messaging options chain agnostic

dwojno:
In this PR I've added messaging option configurable same as chain and cacheServer. I've also updated readme, to new changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,28 +39,95 @@ Prerelease version
 $ npm i iam-client-lib@canary
 ```
 
-### Sample Import (TypeScript)
+### Sample Config for browsers (TypeScript)
 
-``` sh
-import { IAM, CacheServerClient } from 'iam-client-lib';
+```js
+import {
+  IAM,
+  WalletProvider,
+  setCacheClientOptions,
+  setChainConfig,
+  setMessagingOptions,
+  MessagingMethod,
+} from 'iam-client-lib'
 
-export class Sample {
+export class App {
     private _iam: IAM;
 
     constructor() {
-      // create default cache server client (optional)
-        const cacheClient = new CacheServerClient({ url: 'http://cache-server.com'})
+      // IAM has builtin default settings for VOLTA CHAIN
+      
+      // If you want to change default cache server config or add config for your network
+      setCacheClientOptions(1111, {
+        url: 'https://some-cache-server.com/',
+        cacheServerSupportsAuth: true,
+      })
 
-      // create IAM instance with provided rpc URL
-        this._iam = new IAM({
-          rpcUrl: 'https://volta-rpc.energyweb.org/',
-          chainId: 73799,
-          cacheClient // optional
-        });
+      // If you want to change default chain config or add config for your network
+      setChainConfig(1111, {
+        didContractAddress: '0x3e2fb24edc3536d655720280b427c91bcb55f3d6',
+        ensRegistryAddress: '0xa372d665f83197a63bbe633ebe19c7bfd4943003',
+        ensResolverAddress: '0xe878bdcf5148307378043bfd2b584909aa48a227',
+        rpcUrl: 'http://some-rpc.com',
+      })
+
+      // If you want to change default messaging config or add config for your network
+      setMessagingOptions(1111, {
+        messagingMethod: MessagingMethod.Nats,
+        natsServerUrl: 'https://some-exchange-server.com',
+      })
+
+      // create IAM instance
+      this._iam = new IAM();
     }
 
     async initializeIAM() {
       // this will show connection modal and authenticate
+      const { did, connected } = await this._iam.initializeConnection({
+        walletProvider: WalletProvider.MetaMask,
+      });
+
+      // after successfully authentication you can retrieve the signer
+      const signer = this._iam.getSigner();
+    }
+
+```
+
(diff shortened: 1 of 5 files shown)


JGiter (approved):

README.md
```diff
@@ -39,28 +39,95 @@ Prerelease version
...
+  MessagingMethod,
+} from 'iam-client-lib'
+
+export class App {
+    private _iam: IAM;
 
-``` sh
-import { IAM, CacheServerClient } from 'iam-client-lib';
+    constructor() {
+      // IAM will be created with default settings for VOLTA CHAIN
```
jrhender, line 58: I don't think is quite the case. I think with Dmitry's changes, we are using whatever chainID we get from the provider: https://github.com/energywebfoundation/iam-client-lib/blob/faeec472c6d9263b984ab85063dc3c66bc9ad0ff/src/iam/iam-base.ts#L620

README.md
```diff
@@ -39,28 +39,95 @@ Prerelease version
...
+      setChainConfig(SOME_CHAIN_ID, {
+        didContractAddress: '0x3e2fb24edc3536d655720280b427c91bcb55f3d6',
+        ensRegistryAddress: '0xa372d665f83197a63bbe633ebe19c7bfd4943003',
+        ensResolverAddress: '0xe878bdcf5148307378043bfd2b584909aa48a227',
+        rpcUrl: 'http://some-rpc.com',
+      })
+
+      // create IAM instance
+      this._iam = new IAM({
+        // only for Node.js env you need to pass rpcUrl in the constructor
```
jrhender, line 123: 👍

src/iam/chainConfig.ts
```diff
@@ -1,18 +1,24 @@
 import { VoltaAddress1056 } from "@ew-did-registry/did-ethr-resolver";
 import { CacheServerClientOptions } from "../cacheServerClient/cacheServerClient";
+import { MessagingMethod } from "../utils/constants";
```
jrhender, line 3: tbh, I'm a confused by the purpose of the `MessagingMethod` enum. It seems like we are listing `CacheServer`, `WebRTC` and `SmartContractStorage`, but we have only implemented `CacheServer`? If so, what's the point of having the other ones?

jrhender, line 3: It's also a little confusing that the option is `CacheServer` but they need to provide a `natsServerUrl`. Should the option be `NATS` instead of `CacheServer`?

src/iam/iam-base.ts
```diff
@@ -39,16 +44,9 @@ import { emptyAddress, MessagingMethod, PUBLIC_KEY, WALLET_PROVIDER } from "../u
...
-  chainId?: number;
   infuraId?: string;
-  ensResolverAddress?: string;
-  ensRegistryAddress?: string;
-  didContractAddress?: string;
   ipfsUrl?: string;
   bridgeUrl?: string;
-  messagingMethod?: MessagingMethod;
-  natsServerUrl?: string;
-  cacheClient?: ICacheServerClient;
```
jrhender, line 51: Nice work 👍

src/iam/iam-base.ts
```diff
@@ -421,17 +404,19 @@ export class IAMBase {
   }
 
   private async setupNATS() {
-    if (this._natsServerUrl) {
+    const { messagingMethod, natsServerUrl } = this._messagingOptions || {};
+    if (natsServerUrl && messagingMethod === MessagingMethod.CacheServer) {
```
jrhender, line 408: If the conditional statement that is checking the MessagingMethod is in here, may we should change this method name to `setupMessaging`. What do you think?

src/iam/iam-base.ts
```diff
@@ -621,17 +606,28 @@ export class IAMBase {
   private async initChain() {
     const { chainId } = await this._provider.getNetwork();
     const { ensRegistryAddress, ensResolverAddress, didContractAddress } = chainConfigs[chainId];
+
+    if (!ensRegistryAddress)
+      throw new Error(`Chain config for chainId: ${chainId} does not contain ensRegistryAddress`);
+    if (!ensResolverAddress)
+      throw new Error(`Chain config for chainId: ${chainId} does not contain ensResolverAddress`);
+    if (!didContractAddress)
+      throw new Error(`Chain config for chainId: ${chainId} does not contain didContractAddress`);
```
jrhender, line 615: Nice ⭐

jrhender (commented):
Nice work on fixing on the docs and aligning the config approach for messaging 💯 
Added some comments regarding clarifying the messaging options, might be good to address those

src/iam/chainConfig.ts
dwojno: Other ones as @manihagh said are going to be implemented in the future.

dwojno: Yeah I think the naming is not correct, cause as I said, cacheServer !== nats server

src/iam/iam-base.ts
dwojno: 👍

src/iam/chainConfig.ts
jrhender: Ok, thanks.

jrhender (approved):
Looks great!

PR #136, opened 2021-03-08
chore(ew-did-registry): update ew-did-registry packages

dwojno:


JGiter (commented):
Changes requested by Piotr are contained in `latest` version which is 0.1.0-alpha.0. Could we update to it?

JGiter (approved):

PR #141, opened 2021-03-10
Fix/myen 625 qr code when logging out

jrhender:
Note the 3 commits, maybe easier to look at them separately.
fix: add window name back for KM login -> This is small one-liner, putting back change that was there before.
fix: add ControllableWalletConnect MYEN-625 -> This is workaround to fix MYEN-625
refactor: encapsulated WalletConnect functionality -> I feel like `iam-base` is rather long so thought it would be good to pull out and encapsulate the WalletConnect stuff

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -30,7 +30,7 @@
     "strictNullChecks": true,
     "strictPropertyInitialization": false,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es5"
+    "target": "es2017"
   },
   "include": ["src/**/*", "ethers/**/*"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -30,7 +30,7 @@
     "strictNullChecks": true,
     "strictPropertyInitialization": false,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es2017"
+    "target": "es6"
   },
   "include": ["src/**/*", "ethers/**/*"]
 }
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -90,8 +90,8 @@ export class IAM extends IAMBase {
   static async isMetamaskExtensionPresent() {
     const provider = (await detectEthereumProvider({ mustBeMetaMask: true })) as
       | {
-          request: any;
-        }
+        request: any;
+      }
       | undefined;
 
     const chainId = (await provider?.request({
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -30,7 +30,7 @@
     "strictNullChecks": true,
     "strictPropertyInitialization": false,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es5"
+    "target": "es2017"
```
jrhender, line 33: https://stackoverflow.com/a/61037142

JGiter (approved):

tsconfig.json
jrhender: I installed `es-check` to check whether or not our existing builds are compatible with es5. 

When running against both `iam-client-lib.esm.js` and `iam-client-lib.js`, both reported that they aren't currently compatible with es5.

```
❯ es-check es5 dist/iam-client-lib.js          
ES-Check: there were 1 ES version matching errors.

          ES-Check Error:
          ----
          · erroring file: dist/iam-client-lib.js
          · error: SyntaxError: The keyword 'let' is reserved (1070:0)
```

```
❯ es-check es5 dist/iam-client-lib.esm.js
ES-Check: there were 1 ES version matching errors.

          ES-Check Error:
          ----
          · erroring file: dist/iam-client-lib.esm.js
          · error: SyntaxError: The keyword 'import' is reserved (1:0)
```

So maybe our existing library isn't compatible with es5 🤔  ?

jrhender: For iam-client-lib to be compatible with es5, would all of the node_modules used by iam-client-lib also need to be es5 compatible?

tsconfig.json
```diff
@@ -30,7 +30,7 @@
     "strictNullChecks": true,
     "strictPropertyInitialization": false,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es5"
+    "target": "es6"
```
jrhender, line 33: I think `es6` more reasonable than `es2017`, seems to cover nearly all browsers: https://caniuse.com/es6

image

tsconfig.json
jrhender: after doing some more checking, I think we're current only compatible with es9 actually. I think this is due to the bundling of the nats client: 
image

PR #146, opened 2021-03-22
feat(assets): add asset smart contract integration

dwojno:


JGiter (approved):

src/iam.ts
```diff
@@ -1493,4 +1494,72 @@ export class IAM extends IAMBase {
...
+  // ### ASSETS ###
+  public async registerAsset() {
+    if (!this._address) {
+      throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
+    }
+    const asset = new OfferableIdentityFactory();
+    await asset.deploy(this._address, this._assetManagerAddress);
+  }
+
+  public async offerAsset({ assetDID, offerTo }: { assetDID: string; offerTo: string }) {
```
jrhender, line 1507: @JGiter It looks like the `OfferableIdentityOperator` isn't being used here: https://github.com/energywebfoundation/ew-did-registry/blob/development/packages/proxyIdentity/src/offerableIdentityOperator.ts 
Does that make sense? Or should the offerableIdentityOperator be used?

jrhender (commented):
Looks good so far. @dwojno I think we should have some simple tests of iam-client-lib being used to offer and accept/reject an asset.

src/iam.ts
JGiter: I think that Operator/Resolver purpose is resolving DID document. It is possible to include `offer` to `OfferableIdentityOperator`, but asset contract still has to be invoked directly since `accept` and `reject` are not applicable to Operator

kim-energyweb (approved):
LGTM

PR #149, opened 2021-03-27
fix: add stopwatch logging of iam init

jrhender:
Would be curious to see what the results are of you running this branch with Switchboard @manihagh . Are you able to run Switchboard locally to test (if we have 15 minutes or so, I could help you with that)? Or maybe we can get it into Switchboard dev environment and then you can test there?

I'm seeing output which points to `setDocument` being the culprit. However, this doesn't completely make sense because metamask also needs to do the same thing...
```
  private async setDocument() {
    if (this._did && this._didSigner) {
      const document = new DIDDocumentFull(
        this._did,
        new Operator(this._didSigner, this._registrySetting)
      );
      await document.create();
      this._document = document;
    }
  }
```

image

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -154,8 +154,9 @@ export class IAM extends IAMBase {
    */
   async initializeConnection({
     walletProvider = this._providerType,
-    reinitializeMetamask
-  }: { walletProvider?: WalletProvider; reinitializeMetamask?: boolean } = {}): Promise<
+    reinitializeMetamask,
+    logStopwatch = false
+  }: { walletProvider?: WalletProvider; reinitializeMetamask?: boolean, logStopwatch?: boolean } = {}): Promise<
     InitializeData
   > {
     const { privateKey } = this._connectionOptions;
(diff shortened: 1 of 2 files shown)


jrhender:
`document.create` takes a long time because it reads the documents from chain:
```
  /**
 * Relevant did should have positive cryptocurrency balance to perform
 * the transaction. Create method saves the public key in smart contract's
 * event, which can be qualified as document creation
 *
 * @param did
 * @param context
 * @returns Promise<boolean>
 */
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  async create(): Promise<boolean> {
    const did = await this.did();
    if (await this.readOwnerPubKey(did)) {
      return true;
    }
    const attribute = DIDAttribute.PublicKey;
    const updateData: IUpdateData = {
      algo: Algorithms.Secp256k1,
      type: PubKeyType.VerificationKey2018,
      encoding: Encoding.HEX,
      value: { publicKey: `0x${this.getPublicKey()}`, tag: KeyTags.OWNER },
    };
    const validity = 10 * 60 * 1000;
    await this.update(did, attribute, updateData, validity);
    return true;
  }
```
```
  async readOwnerPubKey(did: string): Promise<string | undefined> {
    const selector = {
      publicKey: { id: `${did}#${KeyTags.OWNER}` },
    };
    const pk = await this.readAttribute(
      did,
      selector,
    );
    return pk ? ((pk as IPublicKey).publicKeyHex as string).slice(2) : undefined;
  }
```

jrhender:
Issue motivating this logging has been resolved

PR #152, opened 2021-04-05
bug/pub key always read from document

JGiter:
Presence of owner public key will checked from cache server instead of directly from document

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -1,7 +1,7 @@
 import { providers, Signer, utils, errors, Wallet } from "ethers";
 import { ethrReg, Operator, Resolver } from "@ew-did-registry/did-ethr-resolver";
 import { labelhash, namehash } from "../utils/ENS_hash";
-import { IServiceEndpoint, RegistrySettings } from "@ew-did-registry/did-resolver-interface";
+import { IServiceEndpoint, RegistrySettings, KeyTags } from "@ew-did-registry/did-resolver-interface";
 import { Methods } from "@ew-did-registry/did";
 import { DIDDocumentFull } from "@ew-did-registry/did-document";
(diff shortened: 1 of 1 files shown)


src/iam/iam-base.ts
```diff
@@ -362,12 +362,18 @@ export class IAMBase {
...
+      this._document = new DIDDocumentFull(
         this._did,
         new Operator(this._didSigner, this._registrySetting)
       );
-      await document.create();
-      this._document = document;
+      let pubKey: IPublicKey | undefined;
+      if (this._cacheClient) {
+        const cachedDoc = await this._cacheClient.getDidDocument({ did: this._did });
+        pubKey = cachedDoc.publicKey.find((pk) => pk.id.endsWith(KeyTags.OWNER));
```
jrhender, line 372: I think this is okay for now, but I think that best case scenario we might be able to reuse the "logic" from `document.create()`. For instance, it might be better if we could use the ew-did-registry to check if the did document is in acceptable state but just resolve the did document from the cache-server instead of the blockchain. I think in this approach, we are duplicating the knowledge that the Owner PublicKey is what it means to have a created DID Document. (i.e. https://github.com/energywebfoundation/ew-did-registry/blob/088b3d55dc9976ba370665b2472ecb0f06dd5b08/packages/did-ethr-resolver/src/implementations/resolver.ts#L179)
Another idea is that the ew-did-registry could expose an `IDocDocument.isCreated()` function which could optionally accept a resolved DID document (which we would provide from the cache-server) 🤔 

But I think it's okay for now 😄

jrhender (approved):

src/iam/iam-base.ts
JGiter: I would also like to avoid duplication, but if did-lib depend on cache server we will have cyclic dependency: 
did-lib -> cache-server -> iam-lib -> did-lib

Agree on encapsulating checking of creation of document in did-lib as well as other utility functionality. I will add it later

PR #154, opened 2021-04-08
Publishing

JGiter:
- [x] `Document is created` status checked from cache server which speeds up login 10 times
- ...


jrhender:
@JGiter  I don't think we should publish the asset functionality of iam-client-lib to master until it is published in the cache-server (which I don't think it is yet?) and finalized and tested in Switchboard Dev. Or is there some reason to publish to master now?

JGiter:
> @JGiter I don't think we should publish the asset functionality of iam-client-lib to master until it is published in the cache-server (which I don't think it is yet?) and finalized and tested in Switchboard Dev. Or is there some reason to publish to master now?

Only that part of the functionality that has already been tested in the development is published - asset registration and transfer. In addition, the publication includes improvements to speed up the login. But perhaps it is better to defer publication until all asset functionality ready. What do you think?

jrhender:
> > @JGiter I don't think we should publish the asset functionality of iam-client-lib to master until it is published in the cache-server (which I don't think it is yet?) and finalized and tested in Switchboard Dev. Or is there some reason to publish to master now?
> 
> Only that part of the functionality that has already been tested in the development is published - asset registration and transfer. In addition, the publication includes improvements to speed up the login. But perhaps it is better to defer publication until all asset functionality ready. What do you think?

Yes, IMO, it is better to wait until asset functionality is ready as we expected external users of `iam-client-lib` to use volta-identitycache.energyweb.org which does not yet have asset functionality

PR #155, opened 2021-04-08
feature/asset enrollment

JGiter:
- [x] To request asset enrollment signature of `IAM.createClaimRequest` will accept optional `did` parameter by default set to owner `did`. Enrollment preconditions are verified on asset
- [x] Now we need differentiate between claim requester (asset owner) and claim subject (asset). Claims produced by did-lib includes `subject` field defined in `JWT` specification but now this field used not on purpose, but to represent claim type: https://github.com/energywebfoundation/iam-client-lib/blob/eaeb20872eb6624d0c6d8c9789b4a6bb064f3ca4/src/iam.ts#L1312. This PR fixes this flaw. Claim type still can be retrived from token data 
- [x] Claim issuer will verify that claim subject is claim requester or owned by requester
- [x] After approving of enrollment request asset owner adds issued claim to asset document

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -1284,12 +1284,14 @@ export class IAM extends IAMBase {
 
   async createClaimRequest({
     issuer,
-    claim
+    claim,
+    did = this._did
   }: {
     issuer: string[];
     claim: { claimType: string; fields: { key: string; value: string | number }[] };
+    did?: string;
   }) {
-    if (!this._did) {
+    if (!did) {
       throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
     }
 
@@ -1298,7 +1300,7 @@ export class IAM extends IAMBase {
         type: ENSNamespaceTypes.Roles,
         namespace: claim.claimType
       }),
-      this.getDidDocument({ includeClaims: true })
+      this.getDidDocument({ did, includeClaims: true })
     ]);
 
     if (!roleDefinition) {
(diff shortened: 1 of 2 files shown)


src/iam.ts
```diff
@@ -1309,17 +1334,17 @@ export class IAM extends IAMBase {
 
     this.verifyEnrolmentPreconditions({ claims: didDocument.service, enrolmentPreconditions });
 
-    const token = await this.createPublicClaim({ data: claim, subject: claim.claimType });
+    const token = await this.createPublicClaim({ data: claim, subject });
```
jrhender, line 1337: 👍  Nice fix

src/iam.ts
```diff
@@ -1332,36 +1357,43 @@ export class IAM extends IAMBase {
...
       throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
     }
 
+    const { sub } = await this.decodeJWTToken({ token }) as { sub: string };
+    if (requester !== sub &&
+      !(await this._cacheClient.getOwnedAssets({ did: requester })).some((a) => a.document.id === sub)
+    ) {
+      throw new Error(ERROR_MESSAGES.CLAIM_REQUESTER_NOT_AUTHORIZED);
+    }
+
```
jrhender, line 1378: @marcin-l-tsh I believe this is the check that is needed for https://energyweb.atlassian.net/browse/SWTCH-809
(does that seem right @JGiter ?)

jrhender (approved):
Looks good 👍

src/iam.ts
JGiter: Right. I will remove it from here then to not add delay to request owned assets

PR #158, opened 2021-04-20
feature/update iam to use new asset contract

JGiter:


src/iam.ts
```diff
@@ -1548,22 +1547,13 @@ export class IAM extends IAMBase {
     if (!this._address) {
       throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
     }
-    const asset = new OfferableIdentityFactory(this._signer);
-    const { address } = await asset.deploy(this._address, this._assetManagerAddress);
-    if (this._cacheClient) {
-      /*
-       * we need to wait until cache server will resolve assets did document
-       * which is taking some time
-       */
```
jrhender, line 1557: Do we no longer need to wait for the cache-server? Do you recall why we needed to do that for the previous implementation?

jrhender (approved):

src/iam.ts
JGiter: My bad. I removed this code because it does not affect asset creation process. I didn't immediately realize that this delay was needed for the user experience. I will put it back

PR #162, opened 2021-04-28
feature/provide list of supported DID methods and their validators

JGiter:
Should be merged after https://github.com/energywebfoundation/iam-client-lib/pull/158, which integrates previous changes of did-lib


jrhender (approved):
Looks good. Please add a small note on usage to the documentation if you have time.

PR #164, opened 2021-04-29
bu/can't sync claim to document

JGiter:
Bug caused by creating `ProxyOperator` for identity `DID` instead of identity address


jrhender (approved):
Looks good

PR #165, opened 2021-04-29
bug/can't publish claim without subject

JGiter:
Bug description: Role claims issued on first iteration of enrollment flow didn't specified role `subject` because it was assumed to be role requester. Now when `subject` is required previously issued claims considered as targeting uncontrolled identities
Fix: Claims without subject will be assumed to target publisher


src/iam.ts
```diff
@@ -350,11 +353,17 @@ export class IAM extends IAMBase {
...
 
-    const { iss, sub, claimData } = (await this.decodeJWTToken({ token })) as {
+    const payload = (await this.decodeJWTToken({ token })) as {
       iss: string;
       sub: string;
       claimData: ClaimData;
     };
+    const { iss, claimData } = payload;
+    let sub = payload.sub;
+    if (!sub || sub.length === 0) {
```
jrhender, line 363: @JGiter I believe that previously, claims had the role namespace as the `sub`, so this is what the code should detect and should replace the `sub` to be `this.did` in this case as well.

example of old incorrect claim:
```
{
  "did": "prequalified.roles.flexmarket.apps.elia.iam.ewc",
  "signer": "did:ethr:0x829b91Fa3e91EA4448365ADA58C7Bad1Ff142866",
  "claimData": {
    "fields": [],
    "claimType": "prequalified.roles.flexmarket.apps.elia.iam.ewc"
  },
  "sub": "prequalified.roles.flexmarket.apps.elia.iam.ewc",
  "iat": 1614851452761,
  "iss": "did:ethr:0x829b91Fa3e91EA4448365ADA58C7Bad1Ff142866"
}
```

jrhender (changes requested):

src/iam.ts
JGiter: Added checking. However, it is strange why the subject was the role

jrhender (approved):
Looks good. Some suggestions though if you think they are worthwhile:
- add tests for various conditions if possible
- add a comment that the whole `if` block is there for legacy reasons, as you explained in your PR comment.

PR #172, opened 2021-05-10
Add asset profiles types for claim data.

DawidGil-jlabs:
Adds types for profiles in data claim, because previously it was any.

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -70,8 +70,24 @@ export type Transaction = {
   from: string;
 };
 
+export interface AssetProfile {
+  name?: string;
+  icon?: string;
+}
+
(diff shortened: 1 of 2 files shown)


src/iam/iam-base.ts
```diff
@@ -70,8 +70,24 @@ export type Transaction = {
...
+  [key:string]: AssetProfile;
+}
+
+export interface Profile {
+  name?: string;
+  birthdate?: string;
+  address?: string;
+  assetProfiles?: AssetProfiles;
+}
+
```
jrhender, line 88: Looks good. Perhaps consider moving these interfaces (and maybe the `ClaimData` interface to a separate file), up to you.

jrhender (approved):

PR #175, opened 2021-05-13
feature/issued claim contains only role info

JGiter:
Issued claim includes only `claimType` and `claimTypeVersion`


jrhender (approved):
LGTM

PR #176, opened 2021-05-15
Swtch 941/logout forbidden

jrhender:
Fixing a bunch of issues with the login flow:

- fix: login as different account issue
  When logout then back in as a separate account was using old cookies.
  So check that if publicKey is there so that if not, redo login to cache-server
- fix: maintain login on reload
  sessionIsActive() should work even if init() hasn't been called
  so moved pubKey retrieval back to constructor
- fix: no sense setting DID if address hasn't been set
  So checking this._address when setDid()
- refactor: use isSessionActive() in loginToCacheServer()
  Better to have one canonical way to check if session is active


DawidGil-jlabs (approved):
Lgtm

PR #177, opened 2021-05-17
fix: scripts for building and testing in windows operating system

DawidGil-jlabs:


package.json
```diff
@@ -15,12 +15,12 @@
...
+    "build:typechain:registry": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ensdomains/ens/build/contracts/ENSRegistry.json",
+    "build:typechain:resolver": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ensdomains/resolver/build/contracts/PublicResolver.json",
+    "build:typechain:offerableIdentity": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ew-did-registry/proxyidentity/build/contracts/OfferableIdentity.json",
+    "build:typechain:identityManager": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ew-did-registry/proxyidentity/build/contracts/IdentityManager.json",
     "prepare": "npm run build",
-    "start-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000",
+    "start-rpc": "ganache-cli -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000",
```
jrhender, line 23: @DawidGil-jlabs I don't think this will work as is, as `run-with-testrpc` both starts `ganache-cli` and then runs the jest command (as you can see in `test:watch` and `test:prod` scripts). So if just using `ganace-cli` then `jest` won't be run.
Why was this change necessary for it work on windows? My understanding is that `run-with-testrpc` is already using `ganache-cli` under the hood.

jrhender: @DawidGil-jlabs Ah, I guess your plan is to use `concurrently` instead? I think one tricky bit if using `concurrently` might be stopping `ganache-cli` after the tests are done (though maybe there is any easy way to do that)

package.json
```diff
@@ -15,12 +15,12 @@
...
-    "build:typechain:resolver": "typechain --target ethers-v4 --outDir ethers './node_modules/@ensdomains/resolver/build/contracts/PublicResolver.json'",
-    "build:typechain:offerableIdentity": "typechain --target ethers-v4 --outDir ethers './node_modules/@ew-did-registry/proxyidentity/build/contracts/OfferableIdentity.json'",
-    "build:typechain:identityManager": "typechain --target ethers-v4 --outDir ethers './node_modules/@ew-did-registry/proxyidentity/build/contracts/IdentityManager.json'",
+    "build:typechain:registry": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ensdomains/ens/build/contracts/ENSRegistry.json",
+    "build:typechain:resolver": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ensdomains/resolver/build/contracts/PublicResolver.json",
+    "build:typechain:offerableIdentity": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ew-did-registry/proxyidentity/build/contracts/OfferableIdentity.json",
+    "build:typechain:identityManager": "typechain --target ethers-v4 --outDir ethers ./node_modules/@ew-did-registry/proxyidentity/build/contracts/IdentityManager.json",
```
jrhender, line 21: Build works fine for me on Mac

package.json
DawidGil-jlabs: Yeah, wanted to use concurrently. However you are right, there might be a problem. So i will remove those scripts and i will leave only fixing building on windows. When i find out a fix for this then i will create new PR.

jrhender: Why is the fix necessary? Does `run-with-testrpc` not work on windows?

DawidGil-jlabs: Yes, there is a bug with it on windows: https://github.com/gnosis/run-with-testrpc/issues/6

DawidGil-jlabs: If You prefer i can create specific scripts for Windows OS, so we do not break actual workflow. Someone on windows machine will need to run 2 scripts instead of one. Linux/Mac users will still run tests with run-with-testrpc. Is it ok for you?

jrhender: I think you can get it to work with `concurrently` (or some other approach) so that the same command works on both Linux/Mac and Windows that would be ideal but if it's easier to have a separate windows script, that's okay for now 👍

DawidGil-jlabs: I updated with additional scripts for windows. I think this have low prio, so i go for easier solution.

JGiter (approved):

PR #181, opened 2021-05-24
[SWTCH-276] implement asset agent delegation

DawidGil-jlabs:
First draft of implementing asset agent delegation - updating asset did document.

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -17,8 +17,7 @@
 
 import { providers, Signer } from "ethers";
 import {
-  Algorithms,
-  DIDAttribute, Encoding,
+  DIDAttribute,
   IDIDDocument,
   IServiceEndpoint,
   IUpdateData,
@@ -57,8 +56,6 @@ import { DIDDocumentFull } from '@ew-did-registry/did-document';
 import { Methods } from '@ew-did-registry/did';
 import { addressOf } from '@ew-did-registry/did-ethr-resolver';
 import { isValidDID } from './utils/did';
-import { Keys } from '@ew-did-registry/keys';
-import { IAttributePayload } from '@ew-did-registry/did-resolver-interface/src/models/operator';
 
 export type InitializeData = {
   did: string | undefined;
@@ -284,17 +281,8 @@ export class IAM extends IAMBase {
       return Boolean(updated);
     }
 
-    const defaultValue: IAttributePayload = { publicKey: `0x${new Keys().publicKey}` };
-
-    const updateData: IUpdateData = {
-      algo: Algorithms.Secp256k1,
-      encoding: Encoding.HEX,
-      ...data,
-      value: { ...defaultValue, ...data.value }
-    };
-
     const operator = new ProxyOperator(this._didSigner, this._registrySetting, addressOf(did));
-    const update = await operator.update(did, DIDAttribute.PublicKey, updateData);
+    const update = await operator.update(did, DIDAttribute.PublicKey, data);
 
     return Boolean(update);
   }
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -270,7 +270,6 @@ export class IAM extends IAMBase {
   }): Promise<boolean> {
     const { didAttribute, data, validity, did } = options;
 
-
     if (!did) {
       if (!this._document) {
         throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -17,7 +17,8 @@
 
 import { providers, Signer } from "ethers";
 import {
-  DIDAttribute,
+  Algorithms,
+  DIDAttribute, Encoding,
   IDIDDocument,
   IServiceEndpoint,
   IUpdateData,
@@ -96,12 +97,11 @@ export enum ENSNamespaceTypes {
  */
 export class IAM extends IAMBase {
   private _subscriptions: Subscription[] = [];
-
   static async isMetamaskExtensionPresent() {
     const provider = (await detectEthereumProvider({ mustBeMetaMask: true })) as
       | {
-      request: any;
-    }
+        request: any;
+      }
       | undefined;
 
     const chainId = (await provider?.request({
@@ -151,9 +151,11 @@ export class IAM extends IAMBase {
    * @returns did string, status of connection and info if the user closed the wallet selection modal
    */
   async initializeConnection({
-                               walletProvider = this._providerType,
-                               reinitializeMetamask
-                             }: { walletProvider?: WalletProvider; reinitializeMetamask?: boolean } = {}): Promise<InitializeData> {
+    walletProvider = this._providerType,
+    reinitializeMetamask
+  }: { walletProvider?: WalletProvider; reinitializeMetamask?: boolean } = {}): Promise<
+    InitializeData
+  > {
     const { privateKey } = this._connectionOptions;
 
     if (!walletProvider && !privateKey) {
@@ -215,9 +217,9 @@ export class IAM extends IAMBase {
    *
    */
   async getDidDocument({
-                         did = this._did,
-                         includeClaims = true
-                       }: { did?: string; includeClaims?: boolean } | undefined = {}) {
+    did = this._did,
+    includeClaims = true
+  }: { did?: string; includeClaims?: boolean } | undefined = {}) {
     if (this._cacheClient && did) {
       try {
         const didDoc = await this._cacheClient.getDidDocument({ did, includeClaims });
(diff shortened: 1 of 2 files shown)


src/iam.ts
```diff
@@ -272,6 +277,27 @@ export class IAM extends IAMBase {
     throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
   }
 
+  async updateAssetDidDocument(options: {
```
JGiter, line 280: Method signature is too vague. I suggest something more descriptive like 
`addAssetToDelegates(assetDid, key?) { ... } `

 `IAM._publicKey` is sensible default value for key

src/iam.ts
```diff
@@ -272,6 +277,27 @@ export class IAM extends IAMBase {
...
+    did: string,
+    value: IAttributePayload
+  }) {
+    const { did, value } = options;
+    if (did && this._didSigner) {
+      const defaultValue: IAttributePayload = { publicKey: `0x${new Keys().publicKey}` };
```
JGiter, line 286: Key added to document must be controlled by asset owner

src/iam.ts
```diff
@@ -272,6 +277,27 @@ export class IAM extends IAMBase {
...
+      const updateData: IUpdateData = {
+        algo: Algorithms.Secp256k1,
+        type: PubKeyType.SignatureAuthentication2018,
+        encoding: Encoding.HEX,
+        value: {...defaultValue, ...value },
+      };
+      const operator = new ProxyOperator(this._didSigner, this._registrySetting, addressOf(did));
+      const update = await operator.update(did, DIDAttribute.PublicKey, updateData);
+      return Boolean(update);
```
JGiter, line 295: Could you please explain what is purpose of returning boolean?

src/iam.ts
```diff
@@ -272,6 +277,27 @@ export class IAM extends IAMBase {
...
+    }
+    throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
```
JGiter, line 297: Its better to separate did and signer checking to simplify debugging

JGiter (changes requested):
Tests would be useful

src/iam.ts
```diff
@@ -272,6 +277,27 @@ export class IAM extends IAMBase {
     throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
   }
 
+  async updateAssetDidDocument(options: {
+    did: string,
+    value: IAttributePayload
```
jrhender, line 282: Yes, I think it would be better if this was more generic and it accepted an `IUpdateData` param.

src/iam.ts
jrhender, line 280: I know I suggested the `updateAssetDidDocument` name but I noticed that the `IAM.publishPublicClaim()` function is shared for both assets and users. I think we should therefore reuse `updateDidDocument()` for assets so that users of `iam-client-lib` will have a more consistent API in the sense there isn't one section that has a separate function for assets and one section that has a shared function.

JGiter: May be I am wrong, but my thinking behind this is that purpose of this PR is to approve asset to act on behalf of its owner. Propsed signature more precisely reflect this and abstract from details of the delegation mechanism (client dosn't need to know about document)

src/iam.ts
DawidGil-jlabs: I was looking on implementation `updateDidDocument` method and creating this one similarly. You think it would be better to return `Promise<utils.BigNumber>` which is returned by update method?

src/iam.ts
jrhender: @JGiter Thanks for elaborating 🙇  That's not quite right though. The purpose is add a publicKey to asset DID document so that the keypair can act **as the asset**.

src/iam.ts
jrhender: I don't think this is the case necessarily. Usually will be but not always.

DawidGil-jlabs: So this one can stay as it is, or should i use `this._publicKey` from iam-base class? I'm a little bit confused with this one.

src/iam.ts
```diff
@@ -262,14 +266,37 @@ export class IAM extends IAMBase {
...
-    throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
+
+    const defaultValue: IAttributePayload = { publicKey: `0x${new Keys().publicKey}` };
+
+    const updateData: IUpdateData = {
+      algo: Algorithms.Secp256k1,
+      encoding: Encoding.HEX,
+      ...data,
+      value: { ...defaultValue, ...data.value }
+    };
```
jrhender, line 294: Why not just using the `IUpdateData` that was passed in as a param?

DawidGil-jlabs: Looking on task description, some of the values should be predefined. For example 

> the key should be of type “sigAuth” 

But yeah, i can remove it.

src/iam.ts
```diff
@@ -262,14 +266,37 @@ export class IAM extends IAMBase {
...
+
+    const operator = new ProxyOperator(this._didSigner, this._registrySetting, addressOf(did));
+    const update = await operator.update(did, DIDAttribute.PublicKey, updateData);
```
jrhender, line 297: @JGiter For `publishPublicClaim()` you instantiated `DIDDocumentFull` with the `ProxyOperator` to make the update. https://github.com/energywebfoundation/iam-client-lib/blob/b0d9446caf581500d084531a4917a2ea05c6d0dc/src/iam.ts#L369 Why is that?

src/iam.ts
jrhender: No, the key is in the `IUpdateData`, passed in as a param

src/iam.ts
jrhender: Ah yea, you're right, my fault. I think though that we can push the setting of those values up to Switchboard Dapp interaction with iam-client-lib (i.e. specify these values in https://energyweb.atlassian.net/browse/SWTCH-895) and keep this function generic.

src/iam.ts
```diff
@@ -1677,4 +1561,135 @@ export class IAM extends IAMBase {
     }
     throw new Error(ERROR_MESSAGES.CACHE_CLIENT_NOT_PROVIDED);
   }
+
+  protected async validateChangeOwnership({
+                                            namespaces,
+                                            newOwner
```
jrhender, line 1567: Minor point but I feel like this indenting is a bit odd. Is this your preferred style?

src/iam.ts
```diff
@@ -262,14 +263,28 @@ export class IAM extends IAMBase {
...
+    did?: string;
     validity?: number;
   }): Promise<boolean> {
-    const { didAttribute, data, validity } = options;
-    if (this._document) {
+    const { didAttribute, data, validity, did } = options;
+
+    if (!this._document) {
+      throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
+    }
```
jrhender, line 273: Looks like you don't need `this._document` if `did` is passed.
Same for `this._didSigner` if `did` is passed

src/iam.ts
```diff
@@ -262,14 +263,28 @@ export class IAM extends IAMBase {
...
+
+    if (!did) {
       const updated = await this._document.update(didAttribute, data, validity);
       return Boolean(updated);
     }
-    throw new Error(ERROR_MESSAGES.DID_DOCUMENT_NOT_INITIALIZED);
+
+    const operator = new ProxyOperator(this._didSigner, this._registrySetting, addressOf(did));
+    const update = await operator.update(did, DIDAttribute.PublicKey, data);
```
jrhender, line 285: I don't think that it will always be a `PublicKey`. Should use the `didAttribute` param

src/iam.ts
DawidGil-jlabs: I will revert that one. I use default from webstorm.

src/iam.ts
DawidGil-jlabs: The problem with not predefined data is that, we have optional parameters in `IUpdateData`. If we do not specify them (those optionals), then we can get an error. Function is still generic, because we use predefined data, but you can override them - i'm merging two objects.

src/iam.ts
DawidGil-jlabs: I need `this._didSigner` when `did` is passed. Without that i can't create ProxyOperator, or can I?

test/assets.testsuite.ts
```diff
@@ -79,4 +81,31 @@ export const assetsTests = () => {
...
+        type: PubKeyType.SignatureAuthentication2018,
+        value: { tag: 'key-1', publicKey: `0x${new Keys().publicKey}` }
+      }
+    });
+    expect(update).toBeTruthy();
+
+    const asset = await iam.getDidDocument({ did: `did:ethr:${assetAddress}` });
+    expect(asset.publicKey.length).toBe(2);
+
+    expect(asset.publicKey[0]).toMatchObject({
```
JGiter, line 106: Order of public keys is not defined by spec. Therefore it is more reliable to check that key contained in list

JGiter (changes requested):

test/assets.testsuite.ts
DawidGil-jlabs: Ok, didn't know that. This test passed multiple times, that's why i left it like that :) I will change that.

test/assets.testsuite.ts
```diff
@@ -79,4 +81,31 @@ export const assetsTests = () => {
     expect(owner).toBe(rootOwner.getAddress());
     expect(offeredTo).toBe(emptyAddress);
   });
+
+  test("update did document for asset", async () => {
+    const assetAddress = await iam.registerAsset();
+
+    const asset1 = await iam.getDidDocument({ did: `did:ethr:${assetAddress}` });
```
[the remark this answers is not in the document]
DawidGil-jlabs: Anyway, did you know that every test have `did:ethr:...` ? At least in assets tests. Should i change them also?

jrhender (approved):
Looks good to me

JGiter (approved):

PR #182, opened 2021-05-28
feature/SWTCH-905 request on-chain role

JGiter:
To distinguish between off-chain and on-chain way of role registration added `RegistrationTypes` enum.
On-chain registration differs from off-chain in:

- agreement claim must be published

`claimTypeVersion` is required in enrollment request, because it is required by on-chain registration

? Verification of enrollment preconditions requires at least one role

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -1395,7 +1395,8 @@ export class IAM extends IAMBase {
       )
     );
 
-    const messageId = Buffer.from("1901", "hex");
+    const typedMsgPrefix = "1901";
+    const messageId = Buffer.from(typedMsgPrefix, "hex");
 
     const agreementHash = solidityKeccak256(
       ["bytes", "bytes32", "bytes32"],
@@ -1434,6 +1435,7 @@ export class IAM extends IAMBase {
     const { claimType: role, claimTypeVersion: version } = claim;
     const token = await this.createPublicClaim({ data: claim, subject });
 
+    // TODO: verfiy onchain
     if (registrationTypes.includes(RegistrationTypes.OffChain)) {
       await this.verifyEnrolmentPreconditions({ subject, role });
     }
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -15,7 +15,7 @@
 // @authors: Kim Honoridez
 // @authors: Daniel Wojno
 
-import { providers, Signer } from "ethers";
+import { providers, Signer, utils } from "ethers";
 import {
   IRoleDefinition,
   IAppDefinition,
@@ -64,6 +64,8 @@ import { addressOf } from "@ew-did-registry/did-ethr-resolver";
 import { isValidDID } from "./utils/did";
 import { chainConfigs } from "./iam/chainConfig";
 
+const { id, keccak256, defaultAbiCoder, solidityKeccak256, arrayify } = utils;
+
 export type InitializeData = {
   did: string | undefined;
   connected: boolean;
@@ -81,6 +83,8 @@ export interface IMessage {
 
 export interface IClaimRequest extends IMessage {
   token: string;
+  registrationTypes: RegistrationTypes[];
+  agreement?: string;
 }
 
 export interface IClaimIssuance extends IMessage {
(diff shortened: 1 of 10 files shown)


src/iam.ts
```diff
@@ -57,6 +57,7 @@ import { DIDDocumentFull } from '@ew-did-registry/did-document';
 import { Methods } from '@ew-did-registry/did';
 import { addressOf } from '@ew-did-registry/did-ethr-resolver';
 import { isValidDID } from './utils/did';
+import { arrayify, defaultAbiCoder, id, keccak256, solidityKeccak256 } from "ethers/utils";
```
jrhender, line 60: I think it's better to import as utils and then reference sub functions: ethers-io/ethers.js#349

jrhender (commented):
Looks like a good start

src/iam.ts
```diff
@@ -1316,58 +1339,90 @@ export class IAM extends IAMBase {
...
 
-    if (!this._natsConnection) {
-      if (this._cacheClient) {
-        return this._cacheClient.requestClaim({ did: subject, message });
+    if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+      if (!version) {
+        throw new Error(ERROR_MESSAGES.ONCHAIN_ROLE_VERSION_NOT_SPECIFIED);
       }
-      throw new NATSConnectionNotEstablishedError();
+      message.agreement = await this.approveRolePublishing({ subject, role, version });
```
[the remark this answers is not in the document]
JGiter: Fixed

src/iam.ts
JGiter: Fixed

src/iam.ts
```diff
@@ -1357,58 +1380,90 @@ export class IAM extends IAMBase {
...
+    const agreement_type_hash = id("Agreement(address subject,bytes32 role,uint256 version)");
+
+    const domainSeparator = keccak256(
+      defaultAbiCoder.encode(
+        ["bytes32", "bytes32", "bytes32", "uint256", "address"],
+        [erc712_type_hash, id("Claim Manager"), id("1.0"), (await this._provider.getNetwork()).chainId, this._claimManager.address]
+      )
+    );
+
+    const messageId = Buffer.from("1901", "hex");
```
jrhender, line 1398: maybe assign `"1901"` to a variable so that it is less of a magic string

src/iam.ts
```diff
@@ -1357,58 +1380,90 @@ export class IAM extends IAMBase {
...
-      this.getDefinition({
-        type: ENSNamespaceTypes.Roles,
-        namespace: claim.claimType
-      }),
-      this.getDidDocument({ did: subject, includeClaims: true })
-    ]);
-
-    if (!roleDefinition) {
-      throw new Error(ERROR_MESSAGES.ROLE_NOT_EXISTS);
+    if (registrationTypes.includes(RegistrationTypes.OffChain)) {
```
jrhender, line 1437: This is probably okay for now, as the `ClaimManager` contract will eventually do it, but maybe we should verify their on-chain preconditions prior to them requesting as well.

src/iam.ts
```diff
@@ -1337,17 +1346,31 @@ export class IAM extends IAMBase {
 
   // NATS
 
-  private verifyEnrolmentPreconditions({
-    claims,
-    enrolmentPreconditions
+  private async verifyEnrolmentPreconditions({
```
jrhender, line 1349: Nice refactoring of this method 👍

src/iam/chainConfig.ts
```diff
@@ -36,7 +37,8 @@ export const chainConfigs: Record<number, ChainConfig> = {
     ensPublicResolverAddress: VOLTA_PUBLIC_RESOLVER_ADDRESS,
     domainNotifierAddress: VOLTA_DOMAIN_NOTIFER_ADDRESS,
     assetManagerAddress: "0xE258fA7D1cc8964D0dEB7204Df947bCa42b2c940",
-    didContractAddress: VoltaAddress1056
+    didContractAddress: VoltaAddress1056,
+    claimManagerAddress: "0x2F259e307D0Ba78902391c070e7b4aA043E74DBB"
```
jrhender, line 41: Maybe we should export the `claimManagerAddress` from `iam-contracts` as I've done with the `DOMAIN_NOTIFIER` and `RESOLVER` addresses. I think this is fine for now though 👍

jrhender (approved):
Some comments but looks good overall

PR #189, opened 2021-06-11
feature/SWTCH-909 issue on-chain role

JGiter:
https://energyweb.atlassian.net/browse/SWTCH-909

`issuer` parameter of `IAM.requestClaimRequest` changed to optional, because at the moment of requesting there are might not be issuers. To persist request it will be published to empty account channel
Had to set ganache network id to 1, because ganache sets its chain id to network id and it breaks tests

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -88,7 +88,9 @@ export interface IClaimRequest extends IMessage {
 }
 
 export interface IClaimIssuance extends IMessage {
+  // issuedToken is is only provided in the case of off-chain role
   issuedToken?: string;
+  // onChainProof is only provided in case of on-chain role
   onChainProof?: string;
   acceptedBy: string;
 }
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -1417,7 +1417,7 @@ export class IAM extends IAMBase {
     ));
   }
 
-  async createOnChainProof(role: string, version: number, expiry: number, subject: string): Promise<string> {
+  private async createOnChainProof(role: string, version: number, expiry: number, subject: string): Promise<string> {
     if (!this._did) {
       throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
     }
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -1407,7 +1407,7 @@ export class IAM extends IAMBase {
         domainSeparator,
         keccak256(defaultAbiCoder.encode(
           ["bytes32", "address", "bytes32", "uint256"],
-          [agreement_type_hash, subject, namehash(role), version]
+          [agreement_type_hash, addressOf(subject), namehash(role), version]
         ))
       ]
     );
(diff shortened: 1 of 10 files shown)


src/iam.ts
```diff
@@ -88,7 +88,8 @@ export interface IClaimRequest extends IMessage {
 }
 
 export interface IClaimIssuance extends IMessage {
-  issuedToken: string;
+  issuedToken?: string;
+  onChainProof?: string;
```
jrhender, line 92:
```suggestion
  // issuedToken is is only provided in the case of off-chain role
  issuedToken?: string;
  // onChainProof is only provided in case of on-chain role
  onChainProof?: string;
```

src/iam.ts
```diff
@@ -1415,6 +1415,45 @@ export class IAM extends IAMBase {
     ));
   }
 
+  async createOnChainProof(role: string, version: number, expiry: number, subject: string): Promise<string> {
```
jrhender, line 1418: If adding a public method, please add documentation for it

jrhender (approved):
Looks good. If you have time to add a test around `issueClaimRequest` that would be nice. Might be tricky to handle call to nats/cache-client but maybe you can mock the NATS connection to check the `message` (maybe could use Jest mock API https://jestjs.io/docs/mock-function-api)

src/iam.ts
JGiter: Good you noticed. This member should be private

test/claimsTests/enrollmentClaimsTests.ts
```diff
@@ -0,0 +1,171 @@
...
+      return { service: {} };
+    });
+    const mockedCacheClient = {
+      getRoleDefinition,
+      getDidDocument
+    };
+
+    Reflect.set(IAM.prototype, "_natsConnection", mockedNatsConnection);
+    Reflect.set(IAM.prototype, "_cacheClient", mockedCacheClient);
+    Reflect.set(IAM.prototype, "_jsonCodec", jsonCodec);
```
jrhender, line 93: Nice work mocking these dependencies 👍

jrhender (commented):
One comment about perhaps removing the issuers field entirely as I think it may not be needed

test/claimsTests/claims.testSuite.ts
```diff
@@ -0,0 +1,7 @@
+import { enrollmentClaimsTests } from "./enrollmentClaimsTests";
+import { selfsignedClaimsTests } from "./selfsignedClaimsTests";
+
+export const claimsTests = () => {
+  describe("Selfsigned claim tests", selfsignedClaimsTests);
+  describe.only("Enrollment claims tests", enrollmentClaimsTests);
```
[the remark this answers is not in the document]
JGiter: My bad, temporarily turned-off to reduce testing time and forgot to turn it on again

src/iam.ts
```diff
@@ -1421,7 +1468,8 @@ export class IAM extends IAMBase {
     subject,
     registrationTypes = [RegistrationTypes.OffChain]
   }: {
-    issuer: string[];
+    // at the moment of requesting there might be no issuers
+    issuer?: string[];
```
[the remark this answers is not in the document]
JGiter: Also for real-time notification of issuers, if any exist, I can add `issuersOfRole` endpoint

PR #190, opened 2021-06-15
bug/SWTCH-995 owner public key in asset document

JGiter:
https://energyweb.atlassian.net/browse/SWTCH-995

Having owner public key in asset document means that asset represents owner, while asset should represent device or service


jrhender (approved):
Looks good

DawidGil-jlabs (approved):

PR #197, opened 2021-06-21
Bug/swtch 910 - remove rpc url from chainConfigs

hrachocode:


src/iam/chainConfig.ts
```diff
@@ -31,7 +30,6 @@ export interface MessagingOptions {
  */
 export const chainConfigs: Record<number, ChainConfig> = {
   [VOLTA_CHAIN_ID]: {
-    rpcUrl: "https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/",
```
jrhender, line 34: Sorry, the story description wasn't very clear. All we need to do here is replace "https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/" with "https://volta-rpc.com". We don't need to actually remove the `rpcUrl` property 😄

package.json
```diff
@@ -15,9 +15,9 @@
...
     "build:contracts": "npm run build:typechain:registry && npm run build:typechain:offerableIdentity && npm run build:typechain:identityManager",
-    "build:typechain:registry": "typechain --target ethers-v4 --outDir ethers './node_modules/@ensdomains/ens/build/contracts/ENSRegistry.json'",
-    "build:typechain:offerableIdentity": "typechain --target ethers-v4 --outDir ethers './node_modules/@ew-did-registry/proxyidentity/build/contracts/OfferableIdentity.json'",
-    "build:typechain:identityManager": "typechain --target ethers-v4 --outDir ethers './node_modules/@ew-did-registry/proxyidentity/build/contracts/IdentityManager.json'",
+    "build:typechain:registry": "typechain --target ethers-v4 --outDir ethers \"./node_modules/@ensdomains/ens/build/contracts/ENSRegistry.json\"",
+    "build:typechain:offerableIdentity": "typechain --target ethers-v4 --outDir ethers \"./node_modules/@ew-did-registry/proxyidentity/build/contracts/OfferableIdentity.json\"",
+    "build:typechain:identityManager": "typechain --target ethers-v4 --outDir ethers \"./node_modules/@ew-did-registry/proxyidentity/build/contracts/IdentityManager.json\"",
```
jrhender, line 20: Nice fix for windows 👍 . In general though, probably would be good to have a comment in the commit about the rationale. Maybe amend your commit or even better would be to separate it into two commits, once for the windows fix and one for the `rpcUrl` rename

jrhender (changes requested):
Just need to change the rpcUrl, not remove it 😄

src/iam/chainConfig.ts
hrachocode: ah omg:D actually I spent most of the time getting familiar with code. and waste max. 1 hour on just coding :D so no worries, will revert back the rpcUrl config and also will change the URL as specified.

package.json
hrachocode: sure no problem.

src/iam/chainConfig.ts
```diff
@@ -10,7 +10,7 @@ import { MessagingMethod } from "../utils/constants";
 
 const VOLTA_CHAIN_ID = 73799;
 export interface ChainConfig {
-  rpcUrl: string;
+  rpcUrl: string,
```
JGiter, line 13: What is the reason of changing separator?

hrachocode: just a typo :/  fixed with a semicolon, strange that even with a comma it passed  tests and built successfully, maybe interface syntax Is supporting both commas and semicolons

jrhender (approved):
Sorry, gave you the wrong url to change to. I pushed a commit to fix it 😄

PR #198, opened 2021-06-21
Bug/switch 980 Parse DID and fetch Address 

ahmedolaibrahim:
- [ ] Ensures that input did can be valid wether it contain `did:ethr:` or not


jrhender (approved):
Looks good to me, nice work!

test/utilsTests/did.testSuit.ts
```diff
@@ -1,8 +1,20 @@
...
+import { supportedDIDMethods, parseDID } from '../../src/utils/did';
 import { Methods } from '@ew-did-registry/did';
 
 export function didTests() {
   it('Erc1056 method should be supported', () => {
     expect(supportedDIDMethods()).toContain(Methods.Erc1056);
   });
-}
+
+  it("should remove DID from address if present", async () => {
```
JGiter, line 9: As DID is did:method:address, I think more proper test name is "Should extract address from DID". Also `didAddress` is just `did` or `identity`

src/iam.ts
```diff
@@ -1588,8 +1588,6 @@ export class IAM extends IAMBase {
       throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
     }
 
-    this._did = parseDID(this._did);
```
JGiter: There is no need to keep such changes in commit history. If possible please try to reset previous commits

ahmedolaibrahim: Yeah, I understand what you mean and I agree but unfortunately, there is no clean way to do that. I'll be more careful next time.

JGiter (approved):

PR #200, opened 2021-06-23
feature/SWTCH-1034 get apps and orgs with relations

JGiter:
Added query parameters `withApps` and `withRoles` to return owned organizations and application with their relationships.
Should be merged after

### DIFF
--- a/src/cacheServerClient/ICacheServerClient.ts
+++ b/src/cacheServerClient/ICacheServerClient.ts
@@ -25,12 +25,16 @@ export interface ICacheServerClient {
   getOrganizationRoles: ({ namespace }: { namespace: string }) => Promise<IRole[]>;
   getOrganizationsByOwner: ({
     owner,
-    excludeSubOrgs
+    excludeSubOrgs,
+    withApps,
+    withRoles
   }: {
     owner: string;
     excludeSubOrgs: boolean;
+    withApps?: boolean;
+    withRoles?: boolean;
   }) => Promise<IOrganization[]>;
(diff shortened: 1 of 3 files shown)


src/cacheServerClient/ICacheServerClient.ts
```diff
@@ -23,11 +23,11 @@ export interface ICacheServerClient {
...
     getApplicationRoles: ({ namespace }: Pick<ClaimsQueryParams, "namespace">) => Promise<IRole[]>;
     getOrganizationRoles: ({ namespace }: Pick<ClaimsQueryParams, "namespace">) => Promise<IRole[]>;
-    getOrganizationsByOwner: ({
-        owner,
-        excludeSubOrgs,
-    }: Pick<ClaimsQueryParams, "owner" | "excludeSubOrgs">) => Promise<IOrganization[]>;
-    getApplicationsByOwner: ({ owner }: Pick<ClaimsQueryParams, "owner">) => Promise<IApp[]>;
+    getOrganizationsByOwner: (
+        owner: ClaimsQueryParams["owner"],
+        opts?: { withRelations?: boolean },
```
Harasz, line 28: I think it would be better to create an interface with possible options. Current approach can lead to the same issue that we have fixed here https://github.com/energywebfoundation/iam-client-lib/pull/279

JGiter: Named interface would prevent name confusion if it was used in both libraries. This `opts` object is used only in two functions here. I think it would be an overkill to define such simple type in one of the lib and export it into other

PR #201, opened 2021-06-23
feat: using IPublicClaim to issue credential

jrhender:
https://energyweb.atlassian.net/browse/SWTCH-902
Trying to remove second signature for issuing off-chain token


src/iam.ts
```diff
@@ -473,9 +473,15 @@ export class IAM extends IAMBase {
    * @returns return issued token
    *
    */
-  async issuePublicClaim({ token }: { token: string }) {
+  async issuePublicClaim({ token, publicClaim }: { token?: string, publicClaim?: IPublicClaim }) {
```
JGiter, line 476: It looks like this method doesn't needed as all it does just forward call to this._issueClaims.issuePublicClaim

JGiter (changes requested):

src/iam.ts
JGiter: We will also need to clear `token` from user data on cache server when it receive issuance claim

JGiter: Although it can be kept if later will need to change way of issuing public claim

JGiter (approved):

PR #204, opened 2021-07-02
feature/SWTCH-1094 staking

JGiter:
depends on https://github.com/energywebfoundation/iam-contracts/pull/32

Staking API represented by two classes:
`StakingPoolService` - for managing pools
`StakingPool` - abstraction of staking pool

Example of use:
```
const stakingService = await StakingPoolService.init(signer); // can be iam.getSigner();
await stakingService.launchStakingPool({org, ...});
const pool = await stakingPool.getPool(org);
await pool.putStake(amount);
...
```


src/iam-client-lib.ts
```diff
@@ -93,3 +93,5 @@ export {
 
 export { GnosisIam as SafeIam } from "./GnosisIam";
 export * from "./utils/did";
+
```
jrhender, line 96: Minor peeve 😆  but maybe don't need the extra line 
```suggestion
```

src/iam.ts
```diff
@@ -778,7 +778,7 @@ export class IAM extends IAMBase {
     newOwner: string;
     returnSteps?: boolean;
   }) {
-     newOwner = parseDID(newOwner);
+      newOwner = parseDID(newOwner);
     const orgNamespaces = [
```
jrhender, line 782: Fixing strange alignment
```suggestion
    newOwner = parseDID(newOwner);
    const orgNamespaces = [
```

src/staking/index.ts
```diff
@@ -0,0 +1,121 @@
...
+import { ERROR_MESSAGES } from "../errors";
+import { IAM } from "../iam";
+import { chainConfigs } from "../iam/chainConfig";
+import { EncodedCall } from "../iam/iam-base";
+
+const { namehash } = utils;
+/**
+ * @interface
+ */
+export interface IStakable {
```
jrhender, line 21: Some initial description on what the interface is for would be nice if you have time. Maybe something like below?
```suggestion
/**
 * @interface
 * @description Provides functionality to launch and interact with staking pools
 */
export interface IStakable {
```

src/staking/index.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @description implements IStakable on top of `base`
+ * 
+ * @param base - constructor of any IAM-superclass
+ * 
+ * @returns constructor of class extending `IAM` and implementing IStakable
+ */
+export function Stakable<T extends Constructor<IAM>>(base: T): new (...args: ConstructorParameters<typeof IAM>) => IAM & IStakable {
+  return class extends base implements IStakable {
+    private stakingPoolFactory: StakingPoolFactory;
```
jrhender, line 67: It's not immediately obvious to me why this approach of a function that returns a constructor is being used instead of basic JS class inheritance. Is there some pattern that is being implemented here?

test/staking.ts
```diff
@@ -0,0 +1,354 @@
...
+
+export const stakingTests = (): void => {
+  const serviceProvider = Wallet.createRandom().connect(provider);
+  const patron = Wallet.createRandom().connect(provider);
+  const patronDID = `did:${Methods.Erc1056}:${patron.address}`;
+
+  const orgName = "orgname";
+  const domain = `${orgName}.${root}`;
+  let serviceProviderIam: IAM & IStakable;
+  let patronIam: IAM & IStakable;
```
jrhender, line 33: Maybe a basic question here, but is it necessary to use interfaces here? If you had a `Staker` class, could you just do `IAM & Staker`? Removing the interfaces would be nice if we don't need the abstraction at this point, IMO

jrhender (commented):
Nice work. I like the attempt to separate new code to avoid making `IAM` even bigger even if I'm not fully understanding the approach

test/iam.test.ts
```diff
@@ -63,34 +66,38 @@ beforeAll(async () => {
...
-  setCacheClientOptions(1, { url: "" });
+  setCacheClientOptions(chainId, { url: "" });
 
-  iam = await createIam(privateKey);
+  await replenish(rootOwner.address);
+  iam = await createIam(rootOwner.privateKey);
 });
 
+/**
+   * @todo should be refactored because some tests depends on 'create root node'
```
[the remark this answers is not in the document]
JGiter: Right. Ideally test suites should be independent from each other, but for now most of them relay on root and organization nodes created by this and org suites

src/staking/index.ts
JGiter: My goal is eventually to separate ENS, Roles and Asset functionality. Class with all these mixins will have constructor `ENS(Roles(Assets(Stakable(IAM))))`. Using inheritance for this would require creation of classes `Stakable extends IAM`, `Assets extends Stakable`, `Roles extends Assets`, `ENS extends Roles` of which only latter would be used. Although inheritance approach is completely doable, from object oriented design point of view there are no `is` relationship between these classes

test/staking.ts
JGiter: Right. As there are no going to be other implementations of `IStakable` it should be abstract class

JGiter: I am agree that interface isn't the best way to typify staking feature, because interface assumes that there might be different implementations. But I can't figure out how `Staker` class to represent "addition of staking feature", because implementation of `Staker` will be based on IAM

jrhender (approved):
Looks good! We can figure out the best way to organize code afterwards

src/staking/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    (await this.pool.requestWithdraw()).wait();
+  }
+
+  /**
+   * @description pays back stake with accumulated reward. Withdrawn unavailable until withdrawn delay ends
+   * @emits StakingPool.StakeWithdrawn
+   */
+  async withdraw(): Promise<void> {
+    (await this.pool.withdraw()).wait();
+  }
```
jrhender, line 168: Might be nice to have a `canWithdraw()` method. Otherwise, I think the only way for a client to, for example, disable a withdraw button until they can withdraw, is to attempt the `withdraw()` and catch an exception. But maybe @DawidGil-jlabs you can indicate `canWithdraw()` would be helpful or not.

jrhender (approved):
Look good, nice work Dmitry!

src/staking/index.ts
DawidGil-jlabs: Good catch. I need to disable somehow withdraw button. So if that's not a problem for you @JGiter, please add method `canWithdraw()`. I can create method that will ask for canWithdraw every few seconds so we will have actual information on the screen.

PR #208, opened 2021-07-12
feature/SWTCH-1112 separate connection to cache server and did registry from iam ininitialization

JGiter:
- Connecting to cache server and DID registry are moved to `connectToCacheServer` and `connectToDID`. These methods can be invoked after iam initialization or as part of it by setting additional initialization flags `initCacheServer` and `initDID`

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -157,35 +157,15 @@ export class IAMBase {
     walletProvider?: WalletProvider;
   }) {
     await this.initSigner({ walletProvider, initializeMetamask });
+    await this.setAddress();
+    this.setDid();
     await this.initChain();
     this.initEventHandlers();
 
     if (this._runningInBrowser) {
       await this.setupMessaging();
     }
-    if (this._signer) {
-      const fromCacheLogin = await this.loginToCacheServer();
-      this._publicKey = fromCacheLogin?.publicKey ?? this._publicKey;
-      this._identityToken = fromCacheLogin?.identityToken;
 
-      // We need a pubKey to create DID document.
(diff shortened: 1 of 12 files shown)


src/iam/iam-base.ts
```diff
@@ -413,8 +428,8 @@ export class IAMBase {
   }
 
   private setJWT() {
-    if (this._didSigner) {
-      this._jwt = new JWT(this._didSigner);
+    if (this._signer) {
+      this._jwt = new JWT(this._signer);
```
jrhender, line 432: Nice 👍

jrhender (approved):
This looks great, nice work 👍

DawidGil-jlabs (approved):
Lgtm. What i understood i can leave initialization for switchboard as it is, and for patron stake view i just need to put `{ initCacheServer: false }` and that's all.

jrhender:
> Lgtm. What i understood i can leave initialization for switchboard as it is, and for patron stake view i just need to put `{ initCacheServer: false }` and that's all.

Yes, I think that is basically right. I think you put `{ walletProvider: Metamask, initCacheServer: false, initDid: false }` for the patron stake view

DawidGil-jlabs:
> > Lgtm. What i understood i can leave initialization for switchboard as it is, and for patron stake view i just need to put `{ initCacheServer: false }` and that's all.
> 
> Yes, I think that is basically right. I think you put `{ walletProvider: Metamask, initCacheServer: false, initDid: false }` for the patron stake view

Oh ok. I thought `initDid` should be enabled, but i do not have knowledge how this initialization works. Thanks for clarification :)

PR #210, opened 2021-07-16
initialize azure signer

yartman:
i was able to test it in local with iam-client-examples in react-dapp. but i had to disable the setDocument method in iam-base.ts because it returns a general error from @ew-did-registery package without details.


src/iam/iam-base.ts
```diff
@@ -246,6 +253,20 @@ export class IAMBase {
...
+    if (walletProvider === WalletProvider.EKC) {
+      try {
+        // proxyURL should be 
+        await EKC.init({ proxyUrl: proxyURL as string });
+        await EKC.login({ mode: 'popup' });
+      } catch (error) {
+        throw new Error(ERROR_MESSAGES.ERROR_IN_AZURE_PROVIDER);
+      }
+      this._signer = EKC.getSigner() as Signer;
+      this._provider = this._signer.provider as providers.Web3Provider;
```
JGiter, line 265: As `EKC.getSigner()` returns `Signer` these castings not needed

src/iam/iam-base.ts
```diff
@@ -314,22 +335,23 @@ export class IAMBase {
...
-        isMetamask === false &&
+        isMetamask === false && !isEKC &&
           this._walletConnectService.getProvider()?.wc.on("disconnect", eventHandler);
         break;
       }
       case "networkChanged": {
         isMetamask
           ? this._metamaskProvider?.on("networkChanged", eventHandler)
-          : this._walletConnectService.getProvider().wc.on("session_update", eventHandler);
+          : !isEKC && this._walletConnectService.getProvider().wc.on("session_update", eventHandler);
```
JGiter, line 354: Why `EKC` doesn't handle network events? Without this connection will stay close

JGiter (changes requested):

PR #211, opened 2021-07-17
update iam-contracts

JGiter:
Updated iam-contract exposes withdrawal delay and minimal staking period in StakingPool needed to calculate time left until withdrawal and request withdrawal became available

### DIFF
--- a/src/staking/index.ts
+++ b/src/staking/index.ts
@@ -143,7 +143,7 @@ export class StakingPool {
       throw new Error(ERROR_MESSAGES.STAKE_WAS_NOT_PUT);
     }
     const requestAvailableFrom = depositStart.add(await this.pool.minStakingPeriod());
-    const now = new BigNumber(new Date().getTime());
+    const now = await this.now();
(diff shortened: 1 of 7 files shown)


DawidGil-jlabs (approved):
looks good to me.

src/staking/index.ts
```diff
@@ -103,20 +103,22 @@ export class StakingPoolService {
...
   }
 }
 
 /**
  * Abstraction over staking pool smart contract
  */
 export class StakingPool {
-  constructor(private pool: StakingPoolContract) { }
+  private pool: StakingPoolContract;
+  constructor(private patron: Required<Signer>, address: string) {
```
jrhender, line 119: Just curious, why the "Required<>" around the Signer? Does Signer have optional parameters that you want to ensure are set?

jrhender (approved):
Looks good

src/staking/index.ts
JGiter: Correct. In order patron to send transaction he must be connected, but `provider` property of Signer is optional

PR #223, opened 2021-07-30
feature/upgrade ethers to v5

JGiter:


src/iam/iam-base.ts
```diff
@@ -177,11 +188,8 @@ export class IAMBase {
         const { privateKey, rpcUrl } = this._connectionOptions;
 
         if (!this._runningInBrowser) {
-            if (!privateKey) {
-                throw new Error(ERROR_MESSAGES.NO_PRIVATE_KEY);
-            }
-            if (!rpcUrl) {
-                throw new Error(ERROR_MESSAGES.NO_RPC_URL);
+            if (!privateKey || !rpcUrl) {
+                throw new Error(ERROR_MESSAGES.CONNECTION_OPTIONS_IN_NODE_NOT_PROVIDED);
```
ahmedolaibrahim, line 192: I'm thinking that it might probably be best to let the user know what is missing rather making the error message too generic.

hrachocode: @ahmedolaibrahim  I am thinking if this is the case when lib is not executing in browser `!this._runningInBrowser` , then how we are going to send a message to user end? or is there a need to do it in this case?

hrachocode (approved):

src/iam/iam-base.ts
ahmedolaibrahim: @hrachocode so what I am referring to is that the error messages for when `privateKey` is missing should be diff from the error message for when `rpcUrl` is missing. that way the user knows what to change rather than guessing which is missing.

hrachocode: ah, agree then.

JGiter: :+1:

ahmedolaibrahim (approved):

PR #225, opened 2021-08-03
Feature/icl 1/add formatter

hrachocode:


JGiter (changes requested):
Please format commit messages according with conventional commits spec. Otherwise this changes will not trigger publishing from master

JGiter:
Please ignore my previous comment. We don't need this changes bump library version, because its not affects functionality

PR #229, opened 2021-08-10
upgrading nuts version and using specific build for client

hrachocode:


src/iam.ts
```diff
@@ -63,7 +63,7 @@ import {
     proof_type_hash,
     typedMsgPrefix,
 } from "./utils/constants";
-import { Subscription } from "nats.ws";
+import { Subscription } from "../node_modules/nats.ws/lib/src/mod.js";
```
jrhender, line 66: Why the need to go into the full path? Is `Subscription` not exported from the index of `nats.ws`?

jrhender (commented):
@hrachocode Are you now able to make a NATS connection via websockets in node?

src/iam.ts
hrachocode: > @hrachocode Are you now able to make a NATS connection via websockets in node?

by a suggestion from @JGiter  I made a manual testing after upgrading `nats.ws` in iam-client-lib to 1.2.0 version, which supports .cjs format 

I created a role and made an enrollment using new nats.ws version on iam-client-lib and setup switchboard-dapp to look for local nats server url `http://localhost:9222`, and it worked fine.

hrachocode: > Why the need to go into the full path? Is `Subscription` not exported from the index of `nats.ws`?

after upgrading to `nats.ws` version 1.2.0 switchboard started throwing error that it cannot load the nats.js and I need the corresponding loader to handle it, I found this issue https://github.com/nats-io/nats.ws/issues/100 

and to resolve it, nats docs suggesting to use a .cjs version of nats
https://github.com/nats-io/nats.ws/blob/master/developer_notes.md#working-with-react-create-app

JGiter: At the first glance at the task, I decided that the goal is only to get rid of the bundling of nats to the library, but we also need to test that the iam can be used in the node environment. For nats mocks in the claims and staking can be replaced by real client
As for the inability to load `nats.ws/nats.js` we need to clarify from Switchboard team is it configured to bundle ES modules

JGiter (changes requested):

src/iam.ts
hrachocode: done https://github.com/energywebfoundation/iam-client-lib/pull/229/commits/a233ee2357e29287b262e531c5cd5fa396fa7bb8

JGiter:
Closing as duplicate of https://github.com/energywebfoundation/iam-client-lib/pull/272

PR #236, opened 2021-08-19
fix: logging in without tokens. Cache server rejects that, but there is no information for the user about it.

dawidgil:
From switchboard side we do not know if user have 0 tokens when he tries to login. However cache server requires to have some amount of it. The same creation of did - we need some gas.

### DIFF
--- a/src/errors/ErrorMessages.ts
+++ b/src/errors/ErrorMessages.ts
@@ -32,5 +32,4 @@ export enum ERROR_MESSAGES {
     WITHDRAWAL_WAS_NOT_REQUESTED = "Stake withdrawal was not requested",
     STAKE_WAS_NOT_PUT = "Stake was not put",
     INSUFFICIENT_BALANCE = "Signer has insufficient balance",
-    LOGIN_NOT_ENOUGHT_TOKENS = "You are not able to login because you do not have any tokens",
 }
--- a/src/errors/ErrorMessages.ts
+++ b/src/errors/ErrorMessages.ts
@@ -32,4 +32,5 @@ export enum ERROR_MESSAGES {
     WITHDRAWAL_WAS_NOT_REQUESTED = "Stake withdrawal was not requested",
     STAKE_WAS_NOT_PUT = "Stake was not put",
     INSUFFICIENT_BALANCE = "Signer has insufficient balance",
+    LOGIN_NOT_ENOUGHT_TOKENS = "You are not able to login because you do not have any tokens",
 }
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -175,9 +175,10 @@ export class IAMBase {
         const gas = await this._provider.estimateGas(this._transactionOverrides);
         const balance =
             (await this._signer?.provider?.getBalance(await this._signer?.getAddress())) || BigNumber.from(0);
+        // multiplier 2 chosen arbitrarily because it is not known how reasonably to choose it
         const fee = gasPrice.mul(gas).mul(2);
         if (balance.lte(fee)) {
-            throw new Error(ERROR_MESSAGES.LOGIN_NOT_ENOUGHT_TOKENS);
+            throw new Error(ERROR_MESSAGES.INSUFFICIENT_BALANCE);
         }
     }
 
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -159,6 +159,7 @@ export class IAMBase {
         await this.setAddress();
         this.setDid();
         await this.initChain();
+        await this.checkBalanceForConnection();
         this.initEventHandlers();
 
         if (this._executionEnvironment === ExecutionEnvironment.BROWSER) {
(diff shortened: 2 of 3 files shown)


ahmedolaibrahim (commented):
Looks good to me. can you add tests for this?

dawidgil:
> Looks good to me. can you add tests for this?

Hmm, i'm not sure how to reproduce this in tests, because i didn't find a test that works with balance equal to 0. If you have any advise how to start with it, then it would be very helpful :)

src/iam/iam-base.ts
```diff
@@ -155,6 +155,7 @@ export class IAMBase {
         await this.setAddress();
         this.setDid();
         await this.initChain();
+        await this.checkBalanceForConnection();
```
JGiter, line 158: This way user without token will be unable to login. Balance should be checked before creating document

src/iam/iam-base.ts
```diff
@@ -165,6 +166,13 @@ export class IAMBase {
         this.setJWT();
     }
 
+    private async checkBalanceForConnection() {
+        const balance = await this._signer?.provider?.getBalance(await this._signer?.getAddress());
+        if (balance?.toString() === "0") {
```
JGiter, line 171: It is not enough to ensure that transaction fee will be paid. Balance should be compared with estimated pay of the transaction

JGiter (changes requested):

src/iam/iam-base.ts
dawidgil: Updated code to calculate fee like in staking.

jrhender:
@dawidgil Is this PR still relevant?

dawidgil:
Yes, this fixes a problem for user when he do not have tokens. Right now there is no information to him, why he can't login.

src/iam/iam-base.ts
```diff
@@ -169,6 +170,17 @@ export class IAMBase {
         this.setJWT();
     }
 
+    private async checkBalanceForConnection(): Promise<void> {
+        const gasPrice = await this._provider.getGasPrice();
+        const gas = await this._provider.estimateGas(this._transactionOverrides);
+        const balance =
+            (await this._signer?.provider?.getBalance(await this._signer?.getAddress())) || BigNumber.from(0);
+        const fee = gasPrice.mul(gas).mul(2);
```
jrhender, line 178: Why `mul(2)`? Just to be on the safe side?

dawidgil: Yes, i did the same calculation as Dmitry did here: https://github.com/energywebfoundation/iam-client-lib/blob/4f2ec9ac05316c89d878a7c35568bd68cc3c5424/src/staking/index.ts#L178

src/errors/ErrorMessages.ts
```diff
@@ -32,4 +32,5 @@ export enum ERROR_MESSAGES {
     WITHDRAWAL_WAS_NOT_REQUESTED = "Stake withdrawal was not requested",
     STAKE_WAS_NOT_PUT = "Stake was not put",
     INSUFFICIENT_BALANCE = "Signer has insufficient balance",
+    LOGIN_NOT_ENOUGHT_TOKENS = "You are not able to login because you do not have any tokens",
```
JGiter, line 35: Why not to use `INSUFFICIENT_BALANCE`? If needed you can extend it with failed operation name

src/iam/iam-base.ts
jrhender: Ah gotcha. I think it would be good to have the same comment then
```suggestion
        // multiplier 2 chosen arbitrarily because it is not known how reasonably to choose it (as done in iam-client-lib)
        const fee = gasPrice.mul(gas).mul(2);
```

JGiter (commented):
Since there are a several signing operation in the library I suggest to extract check balance

src/errors/ErrorMessages.ts
dawidgil: I thought it is more to actions like staking. I can change this and i will remove added message.

jrhender (approved):
Looks good to me

src/errors/ErrorMessages.ts
JGiter: This only checks that balance isn't empty, which still doesn't prevent insufficient funds error. @jrhender wouldn't it make sense to verify balance sufficiency for invoking `Operator` methods cost in did-reg?

jrhender: @JGiter I don't understand, isn't there a check for the required funds amount in this PR? E.g.
```
private async checkBalanceForConnection(): Promise<void> {
        const gasPrice = await this._provider.getGasPrice();
        const gas = await this._provider.estimateGas(this._transactionOverrides);
        const balance =
            (await this._signer?.provider?.getBalance(await this._signer?.getAddress())) || BigNumber.from(0);
        // multiplier 2 chosen arbitrarily because it is not known how reasonably to choose it
        const fee = gasPrice.mul(gas).mul(2);
        if (balance.lte(fee)) {
            throw new Error(ERROR_MESSAGES.INSUFFICIENT_BALANCE);
        }
    }
```
It might make sense to do it in DID registry. Do you mean having a method like `checkForSufficientBalance()` on the `Operator`?

JGiter: This method can't check transaction sent by did-registry. Possible approach to this is to check balance in `Operator._sendTransaction()` and throw specific error which would caught by iam-lib and propagated to Switchboard

jrhender: Ah yes, I see what you are saying. Sure, I can make a story for this.

jrhender: https://energyweb.atlassian.net/browse/EDR-77

PR #247, opened 2021-09-06
feat: update did-reg package to fix EDR-36

jrhender:
Issue was that previous version of ethers could not estimate gas


knzeng-e (approved):
I works just fine !!

thank you @jrhender

PR #248, opened 2021-09-07
upgrade did registry

JGiter:
Did registry upgraded to create identity owner from JsonRpcProvider


jrhender (approved):
Looks good to me, thanks!

PR #249, opened 2021-09-10
feature/ICL-84 modules

JGiter:
- [x]  Code of IAM and IAMBase classes is split by dedicated services
- [x] Services should be initialized in strict order. First initialized service is SignerService
- [x] migrate tests and docs
- [x] extend documentation

For brief overview of changes look at https://github.com/energywebfoundation/iam-client-lib/blob/feature/ICL-84_modules/README.md#initialization


src/modules/messaging/MessagingService.ts
```diff
@@ -0,0 +1,75 @@
+import { Codec, connect, JSONCodec, NatsConnection, Subscription } from "nats.ws";
```
Harasz, line 1: We should name this file `messaging.service.ts`

src/init.ts
```diff
@@ -0,0 +1,65 @@
...
+import { DomainsService } from "./modules/domains/domains.service";
+import { AssetsService } from "./modules/assets/assets.service";
+
+export async function initWithPrivateKeySigner(privateKey: string, rpcUrl: string) {
+    const signerService = SignerModule.fromPrivateKey(privateKey, rpcUrl);
+    await signerService.onInit();
+    return await init(signerService);
+}
+
+export async function init(signerService: SignerService) {
```
Harasz, line 18: Missing `ClaimsService` initialization

src/modules/signer/privateKeySigner.ts
```diff
@@ -0,0 +1,8 @@
+import { Wallet } from "ethers";
+import { Signer } from "./signer.types";
+
+export class PrivateKeySigner extends Wallet implements Signer {
+    init(): Promise<void> {
+        return new Promise((resolve) => resolve());
```
Harasz, line 6: Are we still working on implementation? 
```suggestion
        return Promise.resolve();
```

src/modules/signer/metamaskSigner.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export class MetaMaskSigner implements Signer {
+    private _signer: providers.JsonRpcSigner;
+
+    get provider() {
+        return this._signer.provider;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    private constructor() {}
```
Harasz, line 47: Can we just remove constructor?

src/modules/signer/walletConnectKms.ts
```diff
@@ -0,0 +1,17 @@
+import WalletConnectProvider from "@walletconnect/web3-provider";
+import { IWalletConnectProviderOptions } from "@walletconnect/types";
+
+export async function connectWithKmsClient(opts: IWalletConnectProviderOptions, ewKeyManagerUrl: string) {
+    const walletConnectProvider = new WalletConnectProvider(opts);
+    walletConnectProvider.wc.on("display_uri", (err, payload) => {
```
Harasz, line 6:
```suggestion
    walletConnectProvider.wc.on("display_uri", (_, payload) => {
```
Just suggestion

Harasz (changes requested):

Harasz:
For overall architecture design its looks good, imo

dawidgil:
For me it looks a lot better! Good job.
However when i tried to check this in switchboard, iam-client-lib do not export a lot of interfaces/enums for example types from cacheClient folder. Could you export all interfaces as well?

src/modules/cacheClient/ICacheClient.ts
```diff
@@ -11,9 +11,9 @@ import {
     IOrganization,
     IRole,
     Order,
-} from "./cacheServerClient.types";
+} from "./cacheClient.types";
 
-export interface ICacheServerClient {
+export interface ICacheClient {
```
jrhender, line 16: Maybe not necessary for the initial refactoring, but I think it might be cool to refactor this interface as well to be by domain. E.g. `IAssetCacheClient`, `IDomainCacheClient`, `IDIDCacheClient`

jrhender (commented):
Sorry for the delay in reviewing, it looks good overall. I think it is a great separation of the code that will make it much easier to work with

src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    async createProofClaim({
+        claimUrl,
+        encryptedSaltedFields,
+    }: {
+        claimUrl: string;
+        encryptedSaltedFields: IProofData;
+    }) {
+        await this._userClaims.createProofClaim(claimUrl, encryptedSaltedFields);
+    }
```
jrhender, line 127: Should these `claim` methods be in the `didRegistry` module?

JGiter: Logically they belong to claims module. I've placed them here because having them in claims module will just duplicate code because claims module doesn't add anything to creation. Also may be `createProofClaim` can be deleted as it is not used anywhere. Also to justify they are presence here we can see them as tokens, which are part of actual claim

jrhender: At the moment, with your current PR, they are also in the `claims` modules. So I think they are already duplicated?
I don't understand why we can't remove them from `didRegistry` module?

I don't understand your point about "seeing them as tokens". What are we seeing as tokens?

Re removing `createProofClaim`, I agree, let's remove it 👍

JGiter: Indeed. It is already in claims service. I `createPublicClaim` is put in claims module and not duplicated in did module, then userClaims will be exported from claims, which seems as internal implementation of did module.
By differentiating between token and claim I mean that value returned from `createPublicClaim` is a part of type `Claim`

jrhender: Ok, thanks, I'm understanding better now, I think it is okay for now 👍

Harasz (approved):

PR #251, opened 2021-09-10
Rename verifyPublicToken param name

Harasz:
BREAKING CHANGE: Rename the prop name from `issuedToken` to `claimUrl` in method `verifyPublicClaim` in `IAM` class.


jrhender (approved):
Looks good, thanks

PR #254, opened 2021-09-13
refactor!: drop support for DID format when offering asset. Use address instead

dawidgil:
While offering asset to other user let's use address instead of did format.

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -1809,9 +1809,8 @@ export class IAM extends IAMBase {
         if (!this._address) {
             throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
         }
-        const [, , offerToAddress] = offerTo.split(":");
         const [, , assetContractAddress] = assetDID.split(":");
-        const tx = this.offerAssetTx({ assetContractAddress, offerTo: offerToAddress });
+        const tx = this.offerAssetTx({ assetContractAddress, offerTo });
         await this.send({
             calls: [tx],
             from: this._address,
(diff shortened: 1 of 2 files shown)


src/iam.ts
```diff
@@ -1809,9 +1809,8 @@ export class IAM extends IAMBase {
         if (!this._address) {
             throw new Error(ERROR_MESSAGES.USER_NOT_LOGGED_IN);
         }
-        const [, , offerToAddress] = offerTo.split(":");
         const [, , assetContractAddress] = assetDID.split(":");
-        const tx = this.offerAssetTx({ assetContractAddress, offerTo: offerToAddress });
+        const tx = this.offerAssetTx({ assetContractAddress, offerTo });
```
JGiter, line 1813: Is this because it forces you to prepend address with `did:ethr`?

JGiter (commented):
What for `!` in commit message type?

src/iam.ts
dawidgil: what i read the `!` in commit message is for Breaking change. Here I read that: https://www.conventionalcommits.org/en/v1.0.0/ Maybe it is not the same solution for commit messages? 🤔 I can remove it if you want.
Yes, we can just use address instead of DID format here like in transferring organizations. It unifies solutions.

JGiter: We are using semantic-release, which requires angular commit message format https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines. It doesn't mention such option

dawidgil: Ok, i will remove it then. I will add a footer with information about breaking change.

JGiter (approved):

PR #257, opened 2021-09-15
feat(createIdentityProof):

knzeng-e:
adding a function to crearte identity proof on delegate
adding unit tests to test creatIdentityProof
adding unit test to test createIdentityProofWithDelegate


src/iam.ts
```diff
@@ -537,6 +538,13 @@ export class IAM extends IAMBase {
         return this._jwt.decode(token);
     }
 
+    /**
+     * createIdentityProof
```
JGiter, line 542: Is this line really needed?

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
...
     }
 
+    /**
+     * createIdentityProofWithDelegate
+     *
+     * @description create a raw identity proof for a delegate
+     * @returns JWT token of created identity
+     *
+     */
+    async createIdentityProofWithDelegate(
```
JGiter, line 567: I think the such signature would be accurate and supplementary to method above
```suggestion
    async createDelegateProof(
```

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
...
+        secp256k1PrivateKey: string,
```
JGiter, line 568:
```suggestion
        delegateKey: string,
```

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
...
+        rpcUrl: string,
+        identityProofDid: string,
```
JGiter, line 570:
```suggestion
        identity: string,
```

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
...
+    ): Promise<string> {
+        const provider = new providers.JsonRpcProvider(rpcUrl);
+        const wallet = new Wallet(secp256k1PrivateKey, provider);
```
JGiter, line 573: Wallet is not really needed to sign. Is it possible to sign just with `jsonwebtoken` npm package?

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
         throw new Error(ERROR_MESSAGES.PROVIDER_NOT_INITIALIZED);
     }
 
+    /**
+     * createIdentityProofWithDelegate
+     *
+     * @description create a raw identity proof for a delegate
```
JGiter, line 563:
```suggestion
     * @description create proof of identity delegate
```

src/iam.ts
```diff
@@ -549,6 +557,34 @@ export class IAM extends IAMBase {
...
+     * @returns JWT token of created identity
```
JGiter, line 564:
```suggestion
     * @returns token of delegate
```

test/utils/createIdentityProof.testSuite.ts
```diff
@@ -0,0 +1,29 @@
+import { IAM } from "../../src/iam";
```
JGiter, line 1: claimsTests seems like a more proper place for this tests

test/utils/createIdentityProof.testSuite.ts
```diff
@@ -0,0 +1,29 @@
...
+import { rpcUrl } from "../setup_contracts";
+
+export const createIdentityProofTests = () => {
+    let rootOwnerIam: IAM;
+
+    beforeAll(async () => {
+        rootOwnerIam = await createIam(rootOwner.privateKey, { createDocument: true });
+    });
+
+    //createIdentityProof Test
```
JGiter, line 12: No real need for this comment

test/utils/createIdentityProof.testSuite.ts
```diff
@@ -0,0 +1,29 @@
...
+    test("Identity token should be defined", async () => {
```
JGiter, line 13: What it actually tests?

test/utils/createIdentityProof.testSuite.ts
```diff
@@ -0,0 +1,29 @@
...
+        const identityToken = await rootOwnerIam.createIdentityProof();
+        expect(identityToken).toBeDefined;
+    });
+
+    //createIdentityProofWithDelegate
+    test("Identity token should be defined for delegate", async () => {
```
JGiter, line 19: I think that more useful would be scenario when delegate would pass `LoginStrategy` authentication on behalf of identity

JGiter (changes requested):

test/utils/createIdentityProof.testSuite.ts
knzeng-e: I Agree

test/utils/createIdentityProof.testSuite.ts
knzeng-e: We finally set this kind of test on passport-did-auth

test/utils/createIdentityProof.testSuite.ts
knzeng-e: Actually, it is just ensuring that the flow don't break, but I agree with you, the value of this test is very limited (just testing it returns something)

package.json
```diff
@@ -69,13 +70,14 @@
...
         "@metamask/detect-provider": "^1.2.0",
-        "@walletconnect/web3-provider": "1.0.0-rc.4",
+        "@walletconnect/web3-provider": "^1.6.5",
         "axios": "^0.21.1",
         "base64url": "^3.0.1",
         "concurrently": "^6.1.0",
         "eth-ens-namehash": "^2.0.8",
         "ethers": "^5.4.4",
         "js-sha3": "^0.8.0",
+        "jsonwebtoken": "^8.5.1",
```
jrhender, line 80: Instead of `jsonwebtoken` package, I think it would be better if we used the JWT package `ew-did-registry`
```suggestion
        "jsonwebtoken": "^8.5.1",
```

test/claimsTests/createIdentityTests.ts
```diff
@@ -0,0 +1,27 @@
...
+    test("Identity token should be defined for delegate", async () => {
+        const ownerAddress = await rootOwner.getAddress();
+        const identityProofDid = `did:ethr:${ownerAddress}`;
+        const delegateIdentityToken = await rootOwnerIam.createDelegateProof(
+            rootOwner.privateKey,
+            rpcUrl,
+            identityProofDid,
+        );
+        expect(delegateIdentityToken).toBeDefined;
+    });
```
jrhender, line 26: I think best case scenario we could also test that these tokens verify, But I think that might be tough without further mocking IPFS. @knzeng-e Before merging, could you `npm link` this with passport-did-auth to check that verification works if you use these methods? Does that make sense?

src/iam.ts
```diff
@@ -559,6 +563,24 @@ export class IAM extends IAMBase {
         throw new Error(ERROR_MESSAGES.PROVIDER_NOT_INITIALIZED);
     }
 
+    /**
+     * @description create a proof of identity delegate
+     * @returns token of delegate
+     */
```
jrhender, line 569: Please document the parameters of the function as well

jrhender (changes requested):
Thanks for the PR Kevin, sorry it took so long to review!

test/claimsTests/createIdentityTests.ts
knzeng-e: if I get you correctly @jrhender, you suggest that I link passport-did-auth and write some tests using those two methods in passport to ensure the verification process ?

JGiter (approved):

jrhender (approved):

PR #258, opened 2021-09-15
refactor(cache): refactor cache client types

Harasz:


Harasz:
I was investigating error 500 on ICS, when I saw in ICL that we are passing wrong params names and TSC wasn't complaining about it (src/cacheServerClient/cacheServerClient.ts line: 243 & 253). So after a discussion with @JGiter we decided to refactor params types in cache client implementation to avoid such error in the future.

jrhender (approved):
Looks good to me, just need to fix the build

JGiter (approved):
Looks good, though there are some overkill from my point of view. For example, parameters wrapping single property like `GetOwnerParam` neither reduces amount of code nor make it more readable

JGiter:
Closed because of https://github.com/energywebfoundation/iam-client-lib/pull/279

PR #267, opened 2021-09-16
feature/ICL-95 optimize tests

JGiter:
https://energyweb.atlassian.net/browse/ICL-95

- [x] replace blockchain event listeners in tests by contract calls
- [x] mock ipfs
- [x] fixed double node migration of changeResolver
- [x] change resolver script takes almost 100 seconds, while it is logic is not compliant with PublicResolver design and it is unclear if we will need this migration at all. Until these questions are resolved this script and corresponding tests are disabled. Should we remove this script?

### DIFF
--- a/test/utils/utils.testSuite.ts
+++ b/test/utils/utils.testSuite.ts
@@ -2,6 +2,7 @@ import { changeResolverTests } from "./changeResolver.testSuite";
 import { didTests } from "./did.testSuit";
 
 export const utilsTests = () => {
-    describe("Change resolver tests", changeResolverTests);
+    // Skipped until either changeResolver script is changed according with Public Resolver design or there will be no sence in migration
+    describe.skip("Change resolver tests", changeResolverTests);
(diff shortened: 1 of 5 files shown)


test/utils/utils.testSuite.ts
```diff
@@ -2,6 +2,7 @@ import { changeResolverTests } from "./changeResolver.testSuite";
 import { didTests } from "./did.testSuit";
 
 export const utilsTests = () => {
-    describe("Change resolver tests", changeResolverTests);
+    // Skipped until either changeResolver script is changed according with Public Resolver designe or there will be no sence in migration
```
jrhender, line 5: Thanks for the comment here. Could you elaborate a bit on what would need to be changed please?

jrhender (approved):
Looks ok to me

test/utils/utils.testSuite.ts
JGiter: Ok. Added comment to `change_resolver`

PR #268, opened 2021-09-17
feature/ICL-78 info about connected wallet chain name and account

Passerino:
The aim was to return necessary info about logged in account. 

Solution:
- account details are returned on Metamask Walled authorization only 
- if other provider then no details returned 

fields returned: 
- chain Name => either Volta or EWT or Unknown as a fallback. The 'Network' refers to 'testnet' or 'mainnet' which is not what we were about to display on front-end
- chain Id 
- account name => couldn't find a way to fetch account name stored in Metamask. Returns hex account number instead
connected wallet => as front-end already posses this info, we won't return it back again

Tests: Metamask integration can be tested using chrome plugin. IMO, the best place to test it is e2e from Front End app.

### DIFF
--- a/docs/api/interfaces/iam_chainConfig.ChainConfig.md
+++ b/docs/api/interfaces/iam_chainConfig.ChainConfig.md
@@ -7,6 +7,7 @@
 ### Properties
 
 - [assetManagerAddress](iam_chainConfig.ChainConfig.md#assetmanageraddress)
+- [chainName](iam_chainConfig.ChainConfig.md#chainname)
 - [claimManagerAddress](iam_chainConfig.ChainConfig.md#claimmanageraddress)
 - [didContractAddress](iam_chainConfig.ChainConfig.md#didcontractaddress)
 - [domainNotifierAddress](iam_chainConfig.ChainConfig.md#domainnotifieraddress)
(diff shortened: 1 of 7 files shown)


src/utils/chain_name_resolver.ts
```diff
@@ -0,0 +1,16 @@
+export enum ChainName {
+    EnergyWebChain = "Energy Web Chain",
+    EnergyWebVoltaTestnet = "Energy Web Volta Testnet",
+    UnknownChain = "Unknown",
+}
+
+export type ChainId = number;
+
+const chains: Map<ChainId, ChainName> = new Map([
```
JGiter, line 9: there is chain configs https://github.com/energywebfoundation/iam-client-lib/blob/f254c16febbf9e28963063460bd5d8344739f383/src/iam/chainConfig.ts#L36, which can be used for this

JGiter (changes requested):

JGiter (commented):
Please format commit messages according with Angular specification https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format. We use them for versioning and changelog generaion

src/utils/chain_name_resolver.ts
Passerino: +1

JGiter (approved):

PR #269, opened 2021-09-20
feature/ICL-96 configure cache server version

JGiter:
Should be merged after https://github.com/energywebfoundation/iam-cache-server/pull/162

### DIFF
--- a/src/cacheServerClient/cacheServerClient.ts
+++ b/src/cacheServerClient/cacheServerClient.ts
@@ -23,6 +23,7 @@ export interface CacheServerClientOptions {
     url: string;
     cacheServerSupportsAuth?: boolean;
 }
+
 export class CacheServerClient implements ICacheServerClient {
     public pubKeyAndIdentityToken: IPubKeyAndIdentityToken | undefined;
     private httpClient: AxiosInstance;
--- a/src/iam/chainConfig.ts
+++ b/src/iam/chainConfig.ts
@@ -49,7 +49,7 @@ export const chainConfigs: Record<number, ChainConfig> = {
 
 export const cacheServerClientOptions: Record<number, CacheServerClientOptions> = {
     [VOLTA_CHAIN_ID]: {
-        url: "https://volta-identitycache.energyweb.org/",
+        url: "https://volta-identitycache.energyweb.org/v1",
         cacheServerSupportsAuth: true,
     },
 };


src/cacheServerClient/cacheServerClient.ts
```diff
@@ -33,9 +36,13 @@ export class CacheServerClient implements ICacheServerClient {
...
 
-    constructor({ url, cacheServerSupportsAuth = true }: CacheServerClientOptions, signer: Signer) {
+    constructor(
+        { url, cacheServerSupportsAuth = true, apiVersion = defaultCacheServerApiVersion }: CacheServerClientOptions,
+        signer: Signer,
+    ) {
+        console.log(`CacheServerClient.constructor url: ${url}, apiVersion: ${apiVersion}`);
         this.httpClient = axios.create({
-            baseURL: url,
+            baseURL: url.endsWith("/") ? `${url}${apiVersion}` : `${url}/${apiVersion}`,
```
jrhender, line 45: I'm just curious what you see as the advantage of having a separate `apiVersion` param over having folks add the version to the `url`?

jrhender (changes requested):
I think the commit should not start with `refactor` as I think this will not trigger a package release (unless we've configured it to do so): https://github.com/semantic-release/commit-analyzer

src/cacheServerClient/cacheServerClient.ts
JGiter: Separation of version from base url would give more flexibility like changing type of versioning, but it is very unlikely be needed. So probably keeping them together is simpler

jrhender: Ok, so is this PR not needed then as we can simply change the `url`?

JGiter:
> I think the commit should not start with `refactor` as I think this will not trigger a package release (unless we've configured it to do so): https://github.com/semantic-release/commit-analyzer

Yes, we configured it already https://github.com/energywebfoundation/iam-client-lib/blob/9a55b466aff5803c6e9e6e9cb2a041e32872a8e3/.releaserc.json#L21

src/cacheServerClient/cacheServerClient.ts
jrhender: Whoops, nevermind, I see you've changed this

jrhender (approved):
Looks good!

PR #276, opened 2021-09-27
fix(nats): change in the way nats is referenced

Passerino:
changed the way nats.ws is referenced according to document https://github.com/nats-io/nats.ws/blob/master/developer_notes.md

### DIFF
--- a/bili.config.ts
+++ b/bili.config.ts
@@ -1,28 +1,20 @@
 import { Config } from "bili";
 
 const config: Config = {
-  plugins: {
-    typescript2: {
-      tsconfigOverride: {
-        include: ["src", "ethers"]
-      }
-    }
-  },
-  extendConfig: config => ({
-    ...config,
-    // The default externals are the package.json dependencies and can only bundle node_modules that are not a dependency.
-    // So need to clear out all externals, to bundle nats.ws package.
-    externals: []
-  }),
-  // nats.ws is bundled because using standard import/require does not work for cjs format.
-  // nat.ws only provides a .mjs file which cannot be required.
-  // include tslib for clients using older version of tslib
-  bundleNodeModules: ["nats.ws", "tslib"],
-  input: "src/iam-client-lib.ts",
-  output: {
-    format: ["cjs", "esm"],
-    minify: false
-  }
+    plugins: {
+        typescript2: {
+            tsconfigOverride: {
+                include: ["src", "ethers"],
+            },
+        },
+    },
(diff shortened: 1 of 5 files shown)


JGiter (commented):
Have your tried if Switchboard runs after this update? I can't login

Passerino:
fixed the problem with front end after updating nats to 1.3

bili.config.ts
```diff
@@ -1,28 +1,28 @@
...
+    },
+    extendConfig: (config) => ({
+        ...config,
+        // The default externals are the package.json dependencies and can only bundle node_modules that are not a dependency.
+        // So need to clear out all externals, to bundle nats.ws package.
+        externals: [],
+    }),
+    // nats.ws is bundled because using standard import/require does not work for cjs format.
+    // nat.ws only provides a .mjs file which cannot be required.
+    // include tslib for clients using older version of tslib
```
jrhender, line 19: This bit about nats.ws can be removed
```suggestion
    // include tslib for clients using older version of tslib
```

src/iam/iam-base.ts
```diff
@@ -25,7 +25,7 @@ import { ClaimManager } from "../../ethers/ClaimManager";
 import { JWT } from "@ew-did-registry/jwt";
 import { ICacheServerClient } from "../cacheServerClient/ICacheServerClient";
 import { detectExecutionEnvironment, ExecutionEnvironment } from "../utils/detectEnvironment";
-import { connect, NatsConnection, JSONCodec, Codec } from "nats.ws";
+import { connect, NatsConnection, Codec, JSONCodec } from "../../node_modules/nats.ws/lib/src/mod.js";
```
jrhender, line 28: Are `Codec` and `JSONCodec` not importable from `nats.ws` index?

jrhender (commented):
@Passerino Have you been able to test iam-client-lib on node?

src/iam/iam-base.ts
Passerino: "The transpilation process seems to not pick up on the right library for nats.ws. To work around the issue, simple insure you import from the cjs file directly" - I confirmed that it also referes to JSONCodes and Codec

JGiter (approved):
Login is working now. Also is it possible to import commonjs build as `nats.ws/nats.js`? And I think we can also use this library in Node.js if we will shim WebSockets as it is shown in https://github.com/nats-io/nats.ws/blob/master/README.md

jrhender (approved):

PR #285, opened 2021-10-01
feat/ICL-109: allow to use private key to login in browser env

dawidgil:
This change allows to use private key to login through frontend application. Current implementation allows to use private key to login only in Node env. This is needed to start e2e testing with private key, to omit metamask login.

### DIFF
--- a/src/iam/iam-base.ts
+++ b/src/iam/iam-base.ts
@@ -82,7 +82,7 @@ export class IAMBase {
     protected _didSigner: EwSigner | undefined;
     protected _identityToken: string | undefined;
     protected _transactionOverrides: utils.Deferrable<providers.TransactionRequest> = {};
-    protected _providerType: WalletProvider | undefined;
+    protected _providerType: WalletProvider;
     protected _publicKey: string | undefined | null;
 
     protected _registrySetting: RegistrySettings;
@@ -144,6 +144,10 @@ export class IAMBase {
             this._publicKey = localStorage.getItem(PUBLIC_KEY);
         }
 
+        if (this._executionEnvironment === ExecutionEnvironment.NODE) {
+            this._providerType = WalletProvider.PrivateKey;
+        }
+
         this._walletConnectService = new WalletConnectService(bridgeUrl, infuraId, ewKeyManagerUrl);
     }
 
@@ -180,15 +184,10 @@ export class IAMBase {
         walletProvider?: WalletProvider;
     }): Promise<AccountInfo | undefined> {
         const { privateKey, rpcUrl } = this._connectionOptions;
-        if (this._executionEnvironment === ExecutionEnvironment.NODE) {
-            if (!privateKey) {
-                throw new Error(ERROR_MESSAGES.PRIVATE_KEY_NOT_PROVIDED);
-            }
-            if (!rpcUrl) {
-                throw new Error(ERROR_MESSAGES.RPC_URL_NOT_PROVIDED);
-            }
-            this._provider = new JsonRpcProvider({ url: rpcUrl });
-            this._signer = new Wallet(privateKey, this._provider);
+
+        if (walletProvider === WalletProvider.PrivateKey) {
+            this.initWithPrivateKey(privateKey, rpcUrl);
+            this._providerType = walletProvider;
             return;
         }
 
@@ -246,6 +245,17 @@ export class IAMBase {
             return;
         }
         throw new Error(ERROR_MESSAGES.WALLET_TYPE_NOT_PROVIDED);
+    }
+
+    private initWithPrivateKey(privateKey, rpcUrl) {
+        if (!privateKey) {
+            throw new Error(ERROR_MESSAGES.PRIVATE_KEY_NOT_PROVIDED);
+        }
+        if (!rpcUrl) {
+            throw new Error(ERROR_MESSAGES.RPC_URL_NOT_PROVIDED);
+        }
+        this._provider = new JsonRpcProvider({ url: rpcUrl });
+        this._signer = new Wallet(privateKey, this._provider);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/iam/iam-base.ts
```diff
@@ -180,7 +180,8 @@ export class IAMBase {
         walletProvider?: WalletProvider;
     }): Promise<AccountInfo | undefined> {
         const { privateKey, rpcUrl } = this._connectionOptions;
-        if (this._executionEnvironment === ExecutionEnvironment.NODE) {
+        const isPrivateKeyInBrowser = this._executionEnvironment === ExecutionEnvironment.BROWSER && privateKey;
```
JGiter, line 183: `isPrivateKeyInBrowser` and `!privateKey` contradicts each other
```suggestion
       if (privateKey) { ... signer from private key}
      if (this._executionEnvironment == ExecutionEnvironment.BROWSER) { throw new Error(ERROR_MESSAGES.PRIVATE_KEY_NOT_PROVIDED)
```

src/iam/iam-base.ts
```diff
@@ -298,7 +299,10 @@ export class IAMBase {
      * @returns boolean that indicates the session state
      */
     public isSessionActive() {
-        return Boolean(this._publicKey) && Boolean(this._providerType);
+        return (
+            (Boolean(this._publicKey) && Boolean(this._providerType)) ||
```
JGiter, line 303: Is this change because of missing provider of type Wallet? If so then it would be more consistent to add this type

JGiter (changes requested):

src/iam/iam-base.ts
```diff
@@ -1,18 +1,18 @@
-import { providers, Signer, utils, Wallet, BigNumber, ethers } from "ethers";
+import { BigNumber, ethers, providers, Signer, utils, Wallet } from "ethers";
```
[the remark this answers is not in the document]
dawidgil: No, i will revert that.

src/iam/iam-base.ts
dawidgil: There is no wallet type which means there is no providerType. What should i put there? Because localStorage is used only by browser and i can't use only publicKey.

src/iam/iam-base.ts
dawidgil: Ok, so you prefer to create below an if statement that uses the same part of code?

JGiter: I mean that if private key is provided, then it should be used to create signer regardless of any other conditions

src/iam/iam-base.ts
JGiter: You can add Wallet providerType. And to avoid storing session in Node.js environment you can check environment before `storeSession`

src/iam/iam-base.ts
```diff
@@ -181,14 +181,13 @@ export class IAMBase {
     }): Promise<AccountInfo | undefined> {
         const { privateKey, rpcUrl } = this._connectionOptions;
         if (this._executionEnvironment === ExecutionEnvironment.NODE) {
```
JGiter, line 183: I think that from now on we don't need to consider execution environment to initialize signer, so you can remove this block

src/iam/iam-base.ts
```diff
@@ -181,14 +181,13 @@ export class IAMBase {
...
-            if (!rpcUrl) {
-                throw new Error(ERROR_MESSAGES.RPC_URL_NOT_PROVIDED);
-            }
-            this._provider = new JsonRpcProvider({ url: rpcUrl });
-            this._signer = new Wallet(privateKey, this._provider);
+            this.initWithPrivateKey(privateKey, rpcUrl);
+            return;
+        }
+
+        if (WalletProvider.PrivateKey === walletProvider) {
```
JGiter, line 188: For consistency
```suggestion
        if (walletProvider === WalletProvider.PrivateKey) {
```

JGiter (changes requested):

src/iam/iam-base.ts
dawidgil: But this will make a breaking change, because Node user will need to set WalletProvider while initializingConnection, am I right?
What's more, for `WalletProvider.PrivateKey` i'm setting `this._providerType = walletProvider;`. Can it also be set for Node env?

JGiter: For now we can set providerType can be set in constructor if node environment is detected

JGiter:
Please stick with conventional-commits spec

src/iam/iam-base.ts
```diff
@@ -340,7 +357,7 @@ export class IAMBase {
                 break;
             }
             default:
-                throw new Error("Event not supported");
+            // throw new Error("Event not supported");
```
JGiter, line 360: Why this change?

test/initializeConnection.testSuite.ts
```diff
@@ -6,10 +6,6 @@ import { WalletProvider } from "../src/types/WalletProvider";
 const iam_withoutKey = new IAM({ rpcUrl });
 
 export const initializeConnectionTests = () => {
-    test("initializeConnection requires privateKey or walletProvider enum", async () => {
```
JGiter, line 9: Why do you think this test is not needed anymore?

src/iam/iam-base.ts
dawidgil: Ah, checking something and I committed this accidentally. Reverted.

test/initializeConnection.testSuite.ts
dawidgil: Because in node env in constructor we are setting providerType always. So it is not possible to have not defined walletProvider.
I was trying to use spyOn this property, but without success.

JGiter: But the test anyway should confirm that error will be thrown when private key is not provided, just with different error message

dawidgil: Ok, i changed error message there.

JGiter (approved):

PR #288, opened 2021-10-06
feature/initialize with EKC signer

JGiter:
- Added `EKC` signer method
- `EKC` module transformed in tests by babel, because module is of ESM type and has `.js` extenstions


yartman:
it fails in test and apparently i can't run tests due to lack of test env

PR #291, opened 2021-10-07
fix(createDelegateProof): set issuer identity

jrhender:
- [x] npm link test in iam-client-example delegate auth script
- [x] npm link test in Switchboard -> Switchboard builds

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -580,7 +580,7 @@ export class IAM extends IAMBase {
             },
         };
         const jwt = new JWT(new Wallet(delegateKey));
-        const identityToken = jwt.sign(payload, { algorithm: "ES256" });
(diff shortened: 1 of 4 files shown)


src/iam.ts
```diff
@@ -580,7 +580,7 @@ export class IAM extends IAMBase {
             },
         };
         const jwt = new JWT(new Wallet(delegateKey));
-        const identityToken = jwt.sign(payload, { algorithm: "ES256" });
+        const identityToken = jwt.sign(payload, { algorithm: "ES256", issuer: identity });
```
knzeng-e, line 583: side question: how is the algorithm type deterministic to the token creation / verification ?
Would a use of "HS256" instead of "ES256" lead to problems ?

knzeng-e (approved):
Thank you John ! All good for me.
I grabbed the rope to ask a question

PR #292, opened 2021-10-08
feat(WalletConnect): change to public WalletConnect relay

Harasz:


JGiter (commented):
Have you managed to connect with this bridge?

Harasz:
I have with Volta RPC, but there is a problem with signing message bc method `eth_sign` is not supported.
On our dev RPC I haven't, dev RPC is not supporting `skipCache` param in request.

JGiter:
I can't get to signing, because after QR code scanning there are endless requests to Volta

src/iam/iam-base.ts
```diff
@@ -242,6 +245,8 @@ export class IAMBase {
             this._provider = new providers.Web3Provider(wcProvider);
             this._signer = this._provider.getSigner();
             this._providerType = walletProvider;
+            const walletConnector = this._walletConnectService.getConnector();
+            this._walletConnectSignMessage = walletConnector.signMessage.bind(walletConnector);
```
JGiter, line 249: There a lot of places in code where we need signing. Wouldn't be better to implement Signer on top of wallet connect?

jrhender (changes requested):
The fix looks good, nice work. It's annoying we needed to use `patch-package` but at least the change isn't very big.

Could you somehow point to the GitHub issue that you created in the code? Maybe squash your commit and add the link to the commit messages? My goal is just to have a way for future devs to know why this was put in and when they can try to remove it.

PR #298, opened 2021-10-12
fix(claims): remove fields from claims

hejkerooo:


src/iam.ts
```diff
@@ -17,12 +17,12 @@
...
 import {
-    IRoleDefinition,
+    DomainReader,
+    EncodedCall,
     IAppDefinition,
     IOrganizationDefinition,
+    IRoleDefinition,
     PreconditionType,
-    EncodedCall,
-    DomainReader,
```
JGiter, line 25: Why this change?

hejkerooo: Sorted imports, eslint rule

test/staking.ts
```diff
@@ -193,7 +193,7 @@ export const stakingTests = (): void => {
 
             const registrationTypes = [RegistrationTypes.OnChain];
             await patronIam.createClaimRequest({
-                claim: { claimType: `${patronRole}.${root}`, claimTypeVersion: 1, fields: [] },
```
jrhender, line 196: I don't understand why this improves the test? Is it because it is unrealistic to have an empty array for the `fields`?

test/claimsTests/enrollmentClaimsTests.ts
```diff
@@ -88,7 +88,7 @@ export function enrollmentClaimsTests() {
     ) {
         const requesterDID = requesterIam.getDid();
         await requesterIam.createClaimRequest({
-            claim: { claimType, claimTypeVersion: version, fields: [] },
```
jrhender, line 91: Same question as below, but I'm not sure how this improves the tests. Does this change the tests so that they would have failed with the previous implementation?
I think what we should actually be doing is adding data to the fields and making sure that it is stripped out when we issue

src/iam.ts
```diff
@@ -1498,7 +1503,7 @@ export class IAM extends IAMBase {
         subject,
         registrationTypes = [RegistrationTypes.OffChain],
     }: {
-        claim: { claimType: string; claimTypeVersion: number; fields: { key: string; value: string | number }[] };
+        claim: { claimType: string; claimTypeVersion: number };
```
jrhender, line 1506: I think in `createClaimRequest` we want to include the `fields` so that the issuer can read them. It is only at the time of issuing that we want to remove the `fields`

test/iam.test.ts
```diff
@@ -1,19 +1,19 @@
 import { BigNumber, utils, Wallet } from "ethers";
```
hejkerooo, line 1: No changes were made for this file - just eslint/prettier

test/staking.ts
```diff
@@ -3,7 +3,7 @@ import { StakingPoolFactory__factory } from "../ethers/factories/StakingPoolFact
 import { RewardPool__factory } from "../ethers/factories/RewardPool__factory";
```
hejkerooo, line 3: No changes were made for this file - just eslint/prettier

PR #299, opened 2021-10-14
feat: persist issued tokens on cache server

Harasz:
Persist all issued token claims on cache server.

### DIFF
--- a/src/cacheServerClient/cacheServerClient.ts
+++ b/src/cacheServerClient/cacheServerClient.ts
@@ -330,4 +330,8 @@ export class CacheServerClient implements ICacheServerClient {
         const { data } = await this.httpClient.get<AssetHistory[]>(`/assets/history/${id}?${query}`);
         return data;
     }
+
+    async saveIssuedToken({ issuedToken }: Pick<ClaimsQueryParams, "issuedToken">) {
+        await this.httpClient.post("/claim/issued", { issuedToken });
+    }
 }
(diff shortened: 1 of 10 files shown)


src/iam.ts
```diff
@@ -487,16 +492,27 @@ export class IAM extends IAMBase {
...
+
+        let issuedToken = "";
+        if (publicClaim) {
+            issuedToken = await this._issuerClaims.issuePublicClaim(publicClaim);
+        }
+        if (token && !publicClaim) {
+            issuedToken = await this._issuerClaims.issuePublicClaim(token);
+        }
+
+        this._cacheClient.saveIssuedToken({ issuedToken });
```
jrhender, line 513: I think this is adding the side-effect of saving the IssuedToken, right? I would add a boolean to the method params, like `saveToken` that is `false` by default that optionally saves the issuedToken

src/cacheServerClient/cacheServerClient.ts
```diff
@@ -330,4 +330,12 @@ export class CacheServerClient implements ICacheServerClient {
...
         return data;
     }
+
+    async saveIssuedToken({ issuedToken }: Pick<ClaimsQueryParams, "issuedToken">) {
+        await this.httpClient.post("/claim/issued", { issuedToken });
+    }
+
+    async getIssuedTokensBySubjects({ subjects }: Pick<ClaimsQueryParams, "subjects">) {
+        await this.httpClient.get(`/claim/issued?subjects=${subjects.join(",")}`);
+    }
```
jrhender, line 340: Prior to merging, please add documentation to methods

Harasz:
@jrhender do we still want that?

PR #304, opened 2021-10-19
chore(dependabot): initial dependabot config

hejkerooo:
@JGiter we need to enable dependabot in settings
@jrhender can you confirm who needs to be notified about pull-requests made from dependabot?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


hejkerooo:
@jrhender I don't think it's possible
https://github.com/dependabot/dependabot-core/issues/2776

There was feature request similar to your question, we can look for alternatives if you want

@JGiter 
It runs on every weekday, Monday to Friday but I think weekly sounds good enough - I will make adjustment

PR #316, opened 2021-10-31
feature/ICL-134 Support of ES256 jwt

JGiter:
- Updated ew-did-registry according with changes in Caused by https://github.com/energywebfoundation/ew-did-registry/pull/350
- `createDelegateProof` creates token of specified JWT algorithm. This will be moved to JWT


.github/workflows/deploy.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Commit to develop branch and merge to master branch
+        if: github.ref == 'refs/heads/master'
```
jrhender, line 67: Odd that this change is a part of this PR? Was it left out before for some reason?

src/errors/ErrorMessages.ts
```diff
@@ -32,4 +32,5 @@ export enum ERROR_MESSAGES {
     WITHDRAWAL_WAS_NOT_REQUESTED = "Stake withdrawal was not requested",
     STAKE_WAS_NOT_PUT = "Stake was not put",
     INSUFFICIENT_BALANCE = "Signer has insufficient balance",
+    JWT_ALGORITHM_NOT_SUPPORTED = "Jwt algorithm no supported",
```
jrhender, line 35:
```suggestion
    JWT_ALGORITHM_NOT_SUPPORTED = "Jwt algorithm not supported",
```

jrhender (commented):
Just one question about the github workflow change

.github/workflows/deploy.yml
JGiter: This error wasn't addressed before, because until now it didn't cause failed builds. I'll move this fix in separate PR

jrhender (approved):
Looks good, thanks!

PR #317, opened 2021-11-01
feat(core): add required fields in role definition

Passerino:
work in progress, tests needed.


src/iam.ts
```diff
@@ -1558,7 +1573,9 @@ export class IAM extends IAMBase {
             throw new Error(ERROR_MESSAGES.SIGNER_NOT_INITIALIZED);
         }
 
-        await this.verifyEnrolmentPrerequisites({ subject, role: claim.claimType });
+        const requiredRoles = await this.verifyEnrolmentPrerequisites({ subject, role: claim.claimType });
+        const requiredRoleDefinitions = await this.getRolesDefinition({ namespaces: requiredRoles });
+        await varifyRequiredParams({ claimParams: claim.claimParams, requiredRoleDefinitions });
```
jrhender, line 1578: I'm a bit confused why this is change is being made. So I guess we were already checking first checking whether or not the subject had the prerequisite/required roles before issuing the claim to then. I think this makes sense as maybe they snuck in a claim request without having the necessary prerequisites.

But then you're adding an additional check that the `claimParams` that I guess were added by the issuer are required by the enrolmentPrequisite roles? I don't think this is right. I think the check should be of the `claimParams` against some `issuerFields` in the requested role metadata.

Sorry, not sure if this is ready for review of not, but thought I would check in

Also, minor typo:
```suggestion
        const requiredRoles = await this.verifyEnrolmentPrerequisites({ subject, role: claim.claimType });
        const requiredRoleDefinitions = await this.getRolesDefinition({ namespaces: requiredRoles });
        await verifyRequiredParams({ claimParams: claim.claimParams, requiredRoleDefinitions });
```

jrhender (commented):
Just some comments 😄

src/iam.ts
```diff
@@ -1421,7 +1423,13 @@ export class IAM extends IAMBase {
...
     }
 
-    private async verifyEnrolmentPrerequisites({ subject, role }: { subject: string; role: string }) {
+    private async verifyEnrolmentPrerequisites({
+        subject,
+        role,
+    }: {
+        subject: string;
+        role: string;
+    }): Promise<string[]> {
```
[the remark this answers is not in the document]
Passerino: thx for this feedback. I agree. The intention was to omit fetching roles twice. I resolved it in different way.

src/iam.ts
Passerino: typo fixed. And as for preconditions - yes , I misunderstood the concept. I changed verification of required params from prerequisites to issued role.

src/iam/roleRequiredFieldsHelper.ts
```diff
@@ -0,0 +1,62 @@
...
+    claimParams,
+    roleDefinition,
+}: {
+    claimParams: Record<string, string> | undefined;
+    roleDefinition: IRoleDefinition;
+}) {
+    if (!roleDefinition) {
+        throw new Error(ERROR_MESSAGES.ROLE_NOT_EXISTS);
+    }
+    if (roleDefinition.metadata && REQUIRED_PARAMS_FIELD_NAME in roleDefinition.metadata) {
```
jrhender, line 52: @Passerino I see now that the story details were a little unclear, sorry about that. It is true that we have a `metadata` property in the `IRoleDefinition` but we also use the string "metadata" to index into the ENS Text resolver. https://github.com/energywebfoundation/iam-contracts/blob/3f3b3edcc39645f5b520ed303a33d0368389eb8d/src/DomainReader.ts#L161
I think, therefore, that it is a bit strange that our `fields` implementation would be different from our `requiredParams` implementation. We currently have our `fields` as a "first-class" property of the `IRoleDefinition` and I think these `issuerFields` should be too. I can see how you ended up putting them in the existing "metadata" of the `IRoleDefinition` but I think we should have another `issuerFields` property of `IRoleDefinition`

PR #319, opened 2021-11-03
bug/master is merged in develop on each PR to develop 

JGiter:


.github/workflows/deploy.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Commit to develop branch and merge to master branch
+        if: github.ref == 'refs/heads/master'
```
jrhender, line 67: I'm a bit confused with the `name` and this `if` check. The name implies that we are committing to `develop` but only if  `github.ref == 'refs/heads/master'`?

jrhender (approved):
Looks great, thanks

PR #320, opened 2021-11-04
feat(core): getters and setters for keys claims and delegates

Passerino:
getters and setters for keys claims and delegates


JGiter (changes requested):
All these changes should be made in didRegistry module

Passerino:
continued in #456

PR #323, opened 2021-11-10
fix(jest): make work of the npm test:watch and test:watch:windows

Passerino:
path to env parameters got fixed


JGiter (approved):
Looks good. Thank you

PR #324, opened 2021-11-15
feature/ICL-66 authorize patron

JGiter:
-  check that patron is enrolled to required role
- verification enrollment is refactored in `isEnrolled`
- BREAKING CHANGE: staking module depends on claims and cache modules in order to verify enrollment


Harasz (approved):

JGiter:
Obsolete because of https://github.com/energywebfoundation/iam-client-lib/blob/develop/src/modules/staking/staking-pool.service.ts

PR #325, opened 2021-11-16
feat: integrate staking-pool

hejkerooo:
- NPM package is not published yet (staking-pool)
- Keeps old staking to make migration easier


e2e/staking-pool.e2e.ts
```diff
@@ -0,0 +1,185 @@
+import { KeyTags } from "@ew-did-registry/did-resolver-interface";
```
JGiter, line 1: Duplication with staking.pool.e2e.ts. can be avoided if combine all tests and include them in two describe blocks, which sets up different staking pool factories

JGiter (changes requested):

e2e/staking-pool.e2e.ts
hejkerooo: IMO there is no point in combining them both as we will remove older staking, old staking code was kept for backward-compatibility purposes. This staking is quite different than previous one.

JGiter (approved):

PR #326, opened 2021-11-16
feat(claims) ICL-135 add issuer fields in role definition and claims

Passerino:
add issuer fields in role definition and claims in new modularised iam-client-lib

### DIFF
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -41,6 +41,7 @@ const version = 1;
 const baseRoleDef = {
     roleType: "org",
     fields: [],
+    issuerFields: [],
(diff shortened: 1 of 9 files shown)


e2e/claims.service.e2e.ts
```diff
@@ -302,7 +303,7 @@ describe("Enrollment claim tests", () => {
         expect(hasRole).toBe(false);
     });
 
-    test("should issue claim request with additional params", async () => {
+    test("should issue claim request with additional issuer fields", async () => {
```
jrhender, line 306: 👍

jrhender (approved):
Looks great, thanks

PR #328, opened 2021-11-17
bug/SWTCH-1250 Switchboard doesn't relogin on changing Metamask account

JGiter:
I can see two ways to enable signer service to listen on events emitted by `Metamask` or `WalletConnect` wallets:
- Underlying signer can extend `ethers.Signer` with `on` method
- wallet provider notifies signer service

I have chosen second way, because extending signer requires to implement `ethers.Signer` instead of getting it from `Web3Provider` wrapper from `ethers`. May be there is a better design

### DIFF
--- a/src/modules/signer/signer.service.ts
+++ b/src/modules/signer/signer.service.ts
@@ -55,11 +55,13 @@ export class SignerService {
     }
 
     async emit(e: ProviderEvent) {
-        for await (const { event, cb } of this._walletEventListeners) {
-            if (event === e) {
-                await cb();
-            }
-        }
+        await Promise.all(
+            this._walletEventListeners
+                .map(({ event, cb }) => {
+                    return e === event ? cb() : null;
+                })
+                .filter(Boolean),
+        );
     }
 
     on(event: ProviderEvent, cb) {
@@ -80,7 +82,7 @@ export class SignerService {
             this.on(ProviderEvent.NetworkChanged, accChangedHandler);
         } else if (this._providerType === ProviderType.WalletConnect) {
             this.on(ProviderEvent.SessionUpdate, accChangedHandler);
-            this.on(ProviderEvent.Disconnected, this.closeConnection());
+            this.on(ProviderEvent.Disconnected, this.closeConnection);
         }
     }
 
--- a/src/modules/signer/signer.service.ts
+++ b/src/modules/signer/signer.service.ts
@@ -1,6 +1,5 @@
 import { BigNumber, providers, utils, Wallet, ethers, Signer } from "ethers";
 import base64url from "base64url";
-import { EventEmitter } from "events";
 import WalletConnectProvider from "@walletconnect/ethereum-provider";
 import { Methods } from "@ew-did-registry/did";
 import { ERROR_MESSAGES } from "../../errors/ErrorMessages";
@@ -21,6 +20,8 @@ export class SignerService {
     private _chainName: string;
 
     private _servicesInitializers: ServiceInitializer[] = [];
+
+    private _walletEventListeners: { event: ProviderEvent; cb: any }[] = [];
 
     constructor(private _signer: Required<Signer>, private _providerType: ProviderType) {}
 
@@ -53,13 +54,16 @@ export class SignerService {
         this._servicesInitializers.push(initializer);
     }
 
-    on(event: ProviderEvent, cb) {
-        const provider = this._signer.provider;
-        if (provider instanceof EventEmitter) {
-            provider.on(event, cb);
-        } else if (provider instanceof WalletConnectProvider) {
-            provider.on(event, cb);
+    async emit(e: ProviderEvent) {
+        for await (const { event, cb } of this._walletEventListeners) {
+            if (event === e) {
+                await cb();
+            }
         }
+    }
+
+    on(event: ProviderEvent, cb) {
+        this._walletEventListeners.push({ event, cb });
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/modules/signer/signer.types.ts
```diff
@@ -8,9 +8,9 @@ export enum ProviderType {
 }
 
 export enum ProviderEvent {
```
jrhender, line 10: I guess these are from the wallet-connect or the metamask code? Are they documented somewhere? If so, could we add a link to where they are documented or defined?

src/modules/signer/signer.service.ts
```diff
@@ -53,15 +54,18 @@ export class SignerService {
         this._servicesInitializers.push(initializer);
     }
 
-    on(event: ProviderEvent, cb) {
-        const provider = this._signer.provider;
-        if (provider instanceof EventEmitter) {
-            provider.on(event, cb);
-        } else if (provider instanceof WalletConnectProvider) {
-            provider.on(event, cb);
+    async emit(e: ProviderEvent) {
```
jrhender, line 57: I'm a bit confused by the naming here. It seems like this more of a `handleEvent` function 🤔 ? It doesn't look like anything is actually being emitted?

src/modules/signer/walletConnectMetamask.ts
```diff
@@ -12,6 +12,8 @@ export const fromWalletConnectMetamask = async (bridge: string, infuraId?: strin
     await walletProvider.enable();
     const provider = new providers.Web3Provider(walletProvider);
     const signerService = new SignerService(provider.getSigner(), ProviderType.MetaMask);
```
jrhender, line 14: This code seems weird. The provider is coming from `createWalletConnectProvider` but `SignerService` is initialized with `ProviderType.MetaMask`? Is this correct?

jrhender (commented):
Just a few comments.

src/modules/signer/signer.service.ts
JGiter: It is emitted in a sense that it notifies registered listeners. It is similar to Node.js EventEmiter.emit

src/modules/signer/walletConnectMetamask.ts
JGiter: thanks for noticing. may be this ProviderType isn't needed really. it is only used to register listeners of connected provider, but it wouldn't harm if listeners of other providers was registered

jrhender: > it is only used to register listeners of connected provider
I don't understand. What is the relationship of this class to WalletConnect or Metamask?

JGiter: This class is notified by wallet to clean session or reinitialize itself when account is changed
 https://github.com/energywebfoundation/iam-client-lib/blob/44e0ad3ecb8c3fecc98bd717e389bbb60cfbbe2e/src/modules/signer/signer.service.ts#L78

src/modules/signer/signer.service.ts
```diff
@@ -71,10 +75,13 @@ export class SignerService {
...
-        this.on(ProviderEvent.AccountChanged, accChangedHandler);
-        this.on(ProviderEvent.NetworkChanged, accChangedHandler);
-        this.on(ProviderEvent.Disconnected, this.closeConnection());
-        this.on(ProviderEvent.SessionUpdate, accChangedHandler);
+        if (this._providerType === ProviderType.MetaMask) {
+            this.on(ProviderEvent.AccountChanged, accChangedHandler);
+            this.on(ProviderEvent.NetworkChanged, accChangedHandler);
+        } else if (this._providerType === ProviderType.WalletConnect) {
+            this.on(ProviderEvent.SessionUpdate, accChangedHandler);
+            this.on(ProviderEvent.Disconnected, this.closeConnection());
```
Harasz, line 83: Shouldn't be like this?
```suggestion
            this.on(ProviderEvent.Disconnected, this.closeConnection);
```

src/modules/signer/signer.service.ts
```diff
@@ -53,15 +54,18 @@ export class SignerService {
...
     }
 
-    on(event: ProviderEvent, cb) {
-        const provider = this._signer.provider;
-        if (provider instanceof EventEmitter) {
-            provider.on(event, cb);
-        } else if (provider instanceof WalletConnectProvider) {
-            provider.on(event, cb);
+    async emit(e: ProviderEvent) {
+        for await (const { event, cb } of this._walletEventListeners) {
```
Harasz, line 58: Could we wrap it to `Promise.all` to increase performance?

src/modules/signer/walletConnectMetamask.ts
```diff
@@ -11,7 +11,9 @@ export const fromWalletConnectMetamask = async (bridge: string, infuraId?: strin
     const walletProvider = createWalletConnectProvider(bridge, infuraId);
     await walletProvider.enable();
     const provider = new providers.Web3Provider(walletProvider);
-    const signerService = new SignerService(provider.getSigner(), ProviderType.MetaMask);
+    const signerService = new SignerService(provider.getSigner(), ProviderType.WalletConnect);
+    walletProvider.on(ProviderEvent.Disconnected, signerService.emit(ProviderEvent.Disconnected));
```
Harasz, line 15:
```suggestion
    walletProvider.on(ProviderEvent.Disconnected, () => signerService.emit(ProviderEvent.Disconnected));
```

src/modules/signer/walletConnectMetamask.ts
```diff
@@ -11,7 +11,9 @@ export const fromWalletConnectMetamask = async (bridge: string, infuraId?: strin
...
+    walletProvider.on(ProviderEvent.SessionUpdate, signerService.emit(ProviderEvent.SessionUpdate));
```
Harasz, line 16:
```suggestion
    walletProvider.on(ProviderEvent.SessionUpdate, () => signerService.emit(ProviderEvent.SessionUpdate));
```

Harasz (changes requested):

Harasz (approved):

jrhender (approved):
Looks good to me!

PR #329, opened 2021-11-17
bug/ICL-145 Forbidden access to assets endpoints

JGiter:
Authentication tests can be added after verification coded is consolidated in ew-did-registry

### DIFF
--- a/src/modules/cacheClient/cacheClient.service.ts
+++ b/src/modules/cacheClient/cacheClient.service.ts
@@ -87,14 +87,18 @@ export class CacheClient implements ICacheClient {
         return Promise.reject(error);
     }
 
-    async testLogin(): Promise<void> {
-        // Simple test to check if logged in or no. TODO: have dedicated endpoint on the cache-server
-        // If receive unauthorized response, expect that refreshToken() will be called
-        await this.getRoleDefinition("testing.if.logged.in");
-    }
-
     async login() {
-        await this.testLogin();
+        const pubKeyAndIdentityToken = await this._signerService.publicKeyAndIdentityToken();
+        const {
+            data: { refreshToken, token },
+        } = await this.httpClient.post<{ token: string; refreshToken: string }>("/login", {
+            identityToken: pubKeyAndIdentityToken.identityToken,
+        });
+        this.refresh_token = refreshToken;
+        if (!this.isBrowser) {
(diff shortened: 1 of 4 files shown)


Harasz (commented):
Will we have a dedicated endpoint on the cache-serve to handle this?

JGiter:
> Will we have a dedicated endpoint on the cache-serve to handle this?

I think that the problem of reusing cookies from previously logged in account is more specific to browser. We could destroy cookies when token doesn't match asset owner, but we don't know if owner or token correspond to actually logged in user

Harasz (approved):

Harasz:
Fair enough, thanks

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -87,14 +87,18 @@ export class CacheClient implements ICacheClient {
...
 
-    async testLogin(): Promise<void> {
-        // Simple test to check if logged in or no. TODO: have dedicated endpoint on the cache-server
-        // If receive unauthorized response, expect that refreshToken() will be called
-        await this.getRoleDefinition("testing.if.logged.in");
-    }
-
     async login() {
-        await this.testLogin();
+        const pubKeyAndIdentityToken = await this._signerService.publicKeyAndIdentityToken();
```
jrhender, line 91: @JGiter Will this mean that the signature is required each user loads `IAM` (e.g. refreshes the page in SB)?

JGiter: Ah, indeed. Will fix this too. Will read it from local storage

jrhender: That is basically what `testLogin` was doing. It was forcing a request so as to be able to control the timing of the signature request.
```
async testLogin(): Promise<void> {
        // Simple test to check if logged in or no. TODO: have dedicated endpoint on the cache-server
        // If receive unauthorized response, expect that refreshToken() will be called
        await this.getRoleDefinition("testing.if.logged.in");
    }
```
The reason for calling it on `login` is that I think it is a better UX if the signature is required when they first load the app/page. Otherwise it could be weird to be asked to sign something at a random point.

JGiter: Make sense. will revert it

PR #331, opened 2021-11-17
bug/ICL-148 Unauthorized to cache server

JGiter:
- fixes bug of not reusing authentication token from previous session
- more cleanly organizes authentication to cache server:
  - `testLogin` is renamed to `isAuthenticated` and will return boolean. It is simpler than catch unauth error in all places where authorization is checked 
  - authorization header for requests in Node.js was set in several places. Now it will be set in constructor only
- allowed empty catch block eslint rule. This is used in `CacheClient.isAuthenticated`

### DIFF
--- a/src/modules/cacheClient/cacheClient.service.ts
+++ b/src/modules/cacheClient/cacheClient.service.ts
@@ -49,7 +49,8 @@ export class CacheClient implements ICacheClient {
     }
 
     /**
-     * Refreshes access token. If login still fails then signs new identity token and requests access token
+     * @description Refreshes access token. If login still fails then signs new identity token and requests access token
+     * After authentication runs previously failed requests
      */
     async handleUnauthenticated() {
         try {
@@ -74,12 +75,8 @@ export class CacheClient implements ICacheClient {
         this.failedRequests = this.failedRequests.filter((callback) => callback());
     }
 
-    addFailedRequest(callback: () => void) {
-        this.failedRequests.push(callback);
-    }
-
     /**
-     * @description if error was returned not from test loging endpoint, then first tries to refresh auth token and if not helps then asks for new
+     * @description Schedules failed requests after login in
      * @param error
      * @returns
      */
@@ -96,7 +93,7 @@ export class CacheClient implements ICacheClient {
             config.url?.indexOf(TEST_LOGIN_ENDPOINT) === -1
         ) {
             const retryOriginalRequest = new Promise((resolve) => {
-                this.addFailedRequest(() => {
+                this.failedRequests.push(() => {
                     resolve(axios(originalRequest));
                 });
             });
--- a/src/modules/cacheClient/cacheClient.service.ts
+++ b/src/modules/cacheClient/cacheClient.service.ts
@@ -12,7 +12,7 @@ import { IPubKeyAndIdentityToken } from "../signer/signer.types";
 import { cacheConfigs } from "../../config/cache.config";
 import { ICacheClient } from "./ICacheClient";
 import { AssetsFilter, ClaimsFilter } from "./cacheClient.types";
-import { SearchType } from ".";
+import { SearchType, TEST_LOGIN_NAMESPACE } from ".";
 
 export class CacheClient implements ICacheClient {
     public pubKeyAndIdentityToken: IPubKeyAndIdentityToken | undefined;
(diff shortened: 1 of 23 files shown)


jrhender (commented):
I think this looks fine, but it's basically reverting to what we had before, so I'm not sure if the `403` bug is fixed?

JGiter:
> I think this looks fine, but it's basically reverting to what we had before, so I'm not sure if the `403` bug is fixed?

You are right. I was confused with authentication logic and missed this point. Will try to refactor it to make more clear

Harasz (approved):

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -35,69 +36,80 @@ export class CacheClient implements ICacheClient {
...
     isAuthEnabled() {
         return this.authEnabled;
     }
 
-    async handleRefreshToken() {
-        const { refreshToken, token } = await this.refreshToken();
+    /**
+     * @description Refreshes access token. If login still fails then signs new identity token and requests access token
+     * After authentication runs previously failed requests
+     */
```
jrhender, line 54: Nice addition of description!

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -35,69 +36,80 @@ export class CacheClient implements ICacheClient {
         });
         this.httpClient.interceptors.response.use((response: AxiosResponse) => {
             return response;
-        }, this.handleUnauthorized.bind(this));
+        }, this.handleError.bind(this));
         this.authEnabled = cacheServerSupportsAuth;
         this.isBrowser = executionEnvironment() === ExecutionEnvironment.BROWSER;
+        if (!this.isBrowser) {
+            this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${this.token}`;
+        }
```
jrhender, line 44: Why was this portion moved to the `init` method? What if the `token` is refreshed? Doesn't the `Authorization` header need to be refreshed as well?
Is `token` even set at this point 🤔

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -281,29 +293,22 @@ export class CacheClient implements ICacheClient {
...
-            }
-            this.pubKeyAndIdentityToken = pubKeyAndIdentityToken;
-            return { token, refreshToken };
+    /**
+     * @description Checks that auth token has been created, has not expired and corresponds to signer
+     * @todo specific endpoint on cache server to return login info instead of error
+     */
+    private async isAuthenticated(): Promise<boolean> {
+        try {
+            await this.getOwnedAssets(this._signerService.did);
```
jrhender, line 308: I think it might be good to actually use the `TEST_LOGIN_ENDPOINT` value here. Otherwise, one might forget to change `TEST_LOGIN_ENDPOINT` when they accomplish the `@todo` of using a dedicated endpoint.

src/modules/cacheClient/cacheClient.service.ts
jrhender: Also, I wonder if we should flag the fact that we are returning the `token` as a possible security vulnerability 🤔

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -281,29 +293,22 @@ export class CacheClient implements ICacheClient {
...
+    }
 
-            if (!this.isBrowser) {
-                this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${token}`;
-                this.refresh_token = refreshToken;
-            }
-            this.pubKeyAndIdentityToken = pubKeyAndIdentityToken;
-            return { token, refreshToken };
+    /**
+     * @description Checks that auth token has been created, has not expired and corresponds to signer
```
jrhender, line 303: I would add to this that "this is done by a request to the server because the auth token is stored an HTTP-only cookie and so the Javascript has no way to check its validity."

jrhender (commented):
Some additional comments. I think there might be some things to address?

src/modules/cacheClient/cacheClient.service.ts
JGiter: It is moved to `init` because including auth header in requests should depend only on environment, which is set only once. You are right about not properly updated auth header value though. I mistakenly supposed that header will be updated along with this.token :man_facepalming: . I suggest to fix it by
```
this.httpClient.defaults.headers.common["Authentication"] = this.authHeader()
```
so that only value of the auth header will be updated, when token is changed, but the http client will not be reconfigured

JGiter: We could mention that tokens, as well as private key and identity token, should not be exposed, though I am not sure how it could help to secure them

PR #334, opened 2021-11-22
bug/ access token is not refreshed

JGiter:


src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -40,7 +40,7 @@ export class CacheClient implements ICacheClient {
         this.authEnabled = cacheServerSupportsAuth;
         this.isBrowser = executionEnvironment() === ExecutionEnvironment.BROWSER;
         if (!this.isBrowser) {
-            this.httpClient.defaults.headers.common["Authorization"] = `Bearer ${this.token}`;
+            this.httpClient.defaults.headers.common["Authorization"] = this.authHeader;
```
jrhender, line 43: I don't understand how this helps. Will `authHeader()` be called on refresh?

JGiter: You are right. It will not work too. Seems there is no simple way to separate configuration of http client from setting of token. Will get back to initial approach then

jrhender (approved):
Looks good to me

PR #358, opened 2021-12-01
update did registry

JGiter:
Changes public claim verification according with new did reg API


jrhender (approved):
Nice, I think this is really good!

PR #361, opened 2021-12-02
feat(claims.service): ICL-159 add hasOnChainRole utility

jrhender:
Implements: https://energyweb.atlassian.net/browse/ICL-159 

@dawidgil should be useful for staking page to check whether or not can try to stake and for seeing whether or not they need to "add" an approve on-chain role to the ClaimManager smart contract (https://energyweb.atlassian.net/browse/SWTCH-1259)

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -63,6 +63,27 @@ export class ClaimsService {
         this._claimManager = chainConfigs()[chainId].claimManagerAddress;
     }
 
+    /**
+     * A utility function to check the blockchain directly if a DID has a role
+     * TODO: fail if the DID chain ID doesn't match the configured signer network connect
+     * @param did The ethr DID to check
+     * @param role The role to check (the full namespace)
+     * @param version The version to check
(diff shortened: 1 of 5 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -63,6 +63,27 @@ export class ClaimsService {
...
+        const data = this._claimManagerInterface.encodeFunctionData("hasRole", [
+            addressOf(did),
+            namehash(role),
+            version,
+        ]);
+        const result = await this._signerService.call({
+            to: this._claimManager,
+            data,
+        });
+        return Boolean(Number.parseInt(result));
```
JGiter, line 84: Isn't `result` boolean? If yes, then Boolean(paseInt(true)) will return false

jrhender: No, result is a hex string. I added a comment to the code, let me know what you think.

JGiter: Its ok then

JGiter (approved):

PR #364, opened 2021-12-03
Feature/ICL-163 register onchain role with claim manager

JGiter:
https://energyweb.atlassian.net/browse/ICL-163

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -32,7 +32,7 @@ import { ClaimData } from "../didRegistry/did.types";
 import { isValidDID } from "../../utils/did";
 import { JWT } from "@ew-did-registry/jwt";
 import { privToPem, KeyType } from "@ew-did-registry/keys";
-import { isIssuedOnchain } from ".";
+import { readyToBeRegisteredOnchain } from "./claims.types";
 
 const { id, keccak256, defaultAbiCoder, solidityKeccak256, namehash, arrayify } = utils;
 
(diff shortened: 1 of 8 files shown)


Harasz (approved):

src/modules/claims/claims.service.ts
```diff
@@ -216,6 +218,38 @@ export class ClaimsService {
         return this._cacheClient.issueClaim(this._signerService.did, message);
     }
 
+    /**
+     * @description Registers issued onchain claim with Claim manager
+     *
+     * @param claimId - id of signed onchain claim
+     */
+    async registerOnchain(claimId: Claim["id"]) {
+        const claim = await this.getClaimById(claimId);
```
jrhender, line 227: I think it is better if it is the `Claim` here so that we don't force another call to the cache-server
```suggestion
    async registerOnchain(claim: Claim) {
```

src/modules/claims/claims.service.ts
```diff
@@ -216,6 +218,38 @@ export class ClaimsService {
...
+        if (!isIssuedOnchain(claim)) {
+            throw new Error(ERROR_MESSAGES.CLAIM_WAS_NOT_ISSUED);
+        }
+        const { token, subjectAgreement, onChainProof } = claim;
+        const { claimData, sub } = this._didRegistry.jwt.decode(token) as {
+            claimData: { claimType: string; claimTypeVersion: number; expiry: number };
```
jrhender, line 233: `expiry` is not needed
```suggestion
            claimData: { claimType: string; claimTypeVersion: number };
```

src/modules/claims/claims.types.ts
```diff
@@ -42,6 +42,35 @@ export interface Claim {
     namespace: string;
 }
 
+export const isIssuedOnchain = (
```
jrhender, line 45: This name makes more sense to me but let me know what you think
```suggestion
export const readyToBeRegisteredOnchain = (
```

jrhender (commented):
Some further comments

src/modules/claims/claims.types.ts
JGiter: Make sense. May be we should also instead of Claim have specialized OnChainClaim and OffchainClaim

jrhender (approved):
Looks good, thanks

PR #365, opened 2021-12-03
feat: add chain name to DID format

Harasz:


e2e/claims.service.e2e.ts
```diff
@@ -25,11 +25,11 @@ const { namehash } = utils;
 
 const provider = new providers.JsonRpcProvider(rpcUrl);
 const staticIssuer = Wallet.createRandom().connect(provider);
-const staticIssuerDID = `did:${Methods.Erc1056}:${staticIssuer.address}`;
+const staticIssuerDID = `did:${Methods.Erc1056}:volta:${staticIssuer.address}`;
```
JGiter, line 28: You can import chain name from https://github.com/energywebfoundation/ew-did-registry/blob/ac62f5913cc71cb443b47c7b38e92f818eb502f9/packages/did/src/models/index.ts#L12

JGiter (approved):

e2e/claims.service.e2e.ts
jrhender: I'll make this change

jrhender (approved):
Looks good to me!

PR #368, opened 2021-12-07
fix(assets): rollback asset event type to uppercase

Passerino:
asset event type is in use in too many places to perform migration easily


jrhender (approved):
Sounds good. maybe not ideal but not that big a deal 😄

PR #375, opened 2021-12-10
fix(nats): fix http to ws replacement

Passerino:
http protocol to ws, https protocol to wss


Harasz (changes requested):
What about `https` and `wss`?

Harasz (approved):
Ohh, I see now. Nice one!

PR #376, opened 2021-12-10
bug/can't register on-chain claim

JGiter:


jrhender (approved):
Makes sense. Nice find! Thanks @JGiter

Harasz (approved):

PR #381, opened 2021-12-13
bug/ICS-124 random unauthorized to cache server

JGiter:
Error with using token of previous user was because result of authentication checking was returned as boolean while re-login is happens on throwing


jrhender (changes requested):
We are not currently getting reports of random authorization currently so I think we should improve our understanding/tests around this auth code before making any further changes as these changes are quite risky https://energyweb.atlassian.net/browse/ICL-170

JGiter:
Closed as duplicate of https://github.com/energywebfoundation/iam-client-lib/pull/556

PR #370, opened 2021-12-07
feature/SWTCH-1268 scripts to create and transfer energyweb

JGiter:
`createEWCTree` creates subdomains according with https://energyweb.atlassian.net/wiki/spaces/EWTS/pages/2960228364/SB+setup. 

1. Set `AUTH_EWC`, `ewOwnerKey` and `CHAIN_ID` in scripts/EW/constants.ts
2. ```ts-node scripts/EW/transferRoot.ts```
3. Set `ENERGYWEB_RECEIVER`
4. ```ts-node scripts/EW/createEWRoot.ts```

- fixed default configuration values

PR #382, opened 2021-12-14
fix: disable staking pool initialization

hejkerooo:


dawidgil (approved):

src/modules/staking/staking.service.ts
```diff
@@ -56,6 +56,11 @@ export class StakingService {
     async init() {
         const chainId = this._signerService.chainId;
         this._stakingPoolFactoryAddress = chainConfigs()[chainId].stakingPoolFactoryAddress;
+
+        if (!this._stakingPoolFactoryAddress || !this._stakingPoolFactoryAddress.length) {
```
JGiter, line 60: Successful initialization of StakingService mean that it can be used, which will lead to errors when stakingPoolFactory isn't set. This should throw an error from my opinion. If we want to disable staking service, then it should be specified explicitly by additional parameter in init functions

JGiter (changes requested):

src/modules/staking/staking.service.ts
hejkerooo: https://github.com/energywebfoundation/iam-client-lib/pull/370/files#diff-85024e9ac487a4d057efb7f1201aba9d16d5ee8d3c8c135a92c9005ce802a858R69

Then this change is incompatible - it's impossible to start any application on EWC right now as the staking address is empty and there is a forced connection to staking pool. IMO it makes sense to not initialize staking when staking pool factory address is empty.

https://github.com/energywebfoundation/iam-client-lib/blob/develop/src/init.ts#L57

JGiter: How about move this check in https://github.com/energywebfoundation/iam-client-lib/blob/5eaa4ec086f7fcdda92c1486688385d06293cd14/src/init.ts#L56 and return null stakingService if it can't be initialized? It allows to detect problems early

src/init.ts
```diff
@@ -49,12 +50,24 @@ export async function init(signerService: SignerService) {
...
+        const stakingPoolFactoryAddress = chainConfigs()[chainId].stakingPoolFactoryAddress;
+
         const cacheClient = new CacheClient(signerService);
         await cacheClient.init();
         await cacheClient.login();
         const domainsService = await DomainsService.create(signerService, cacheClient);
-        const stakingService = await StakingService.create(signerService, domainsService);
-        const stakingPoolService = await StakingFactoryService.create(signerService, domainsService);
+
+        const stakingAddressProvided = stakingPoolFactoryAddress && stakingPoolFactoryAddress.length;
```
jrhender, line 61: This is a test for the stakingPoolFactoryAddress
```suggestion
        const stakingPoolFactoryAddressProvided = stakingPoolFactoryAddress && stakingPoolFactoryAddress.length;
```

src/init.ts
```diff
@@ -49,12 +50,24 @@ export async function init(signerService: SignerService) {
...
+
+        const stakingService = stakingAddressProvided
+            ? await StakingService.create(signerService, domainsService)
+            : null;
+
+        const stakingPoolService = stakingAddressProvided
+            ? await StakingFactoryService.create(signerService, domainsService)
+            : null;
```
jrhender, line 69: @hejkerooo so does the staking dapp not need either of these? I guess we're trying to fix this for KYA, right?

jrhender (approved):
I think the direction of this PR makes sense. Obviously, we can't have initialization of the cache-server depend on needing a `stakingPoolFactory` as we don't have one in prod. Maybe a further refactoring is still needed, but I think let's try this for now 👍

JGiter (approved):

PR #383, opened 2021-12-15
feat(claims): rejection reason while rejecting the claim

Passerino:
rejection reason to give requester more details about the causes

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -278,12 +278,21 @@ export class ClaimsService {
         });
     }
 
-    async rejectClaimRequest({ id, requesterDID }: { id: string; requesterDID: string }) {
+    async rejectClaimRequest({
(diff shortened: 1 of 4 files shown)


jrhender (approved):
Look good to me, thanks

src/modules/claims/claims.service.ts
```diff
@@ -278,12 +278,21 @@ export class ClaimsService {
...
+        id: string;
+        requesterDID: string;
+        rejectionReason?: string;
+    }) {
         const message: IClaimRejection = {
             id,
             requester: requesterDID,
             claimIssuer: [this._signerService.did],
             isRejected: true,
+            rejectionReason,
```
jrhender, line 295: @JGiter This approach makes sense to me, so the cache-server just needs to accept an additional `rejectionReason` string in the rejection message

PR #384, opened 2021-12-15
feature/ICS-170 document cache client authentication

JGiter:
https://energyweb.atlassian.net/browse/ICL-170

### DIFF
--- a/src/modules/cacheClient/cacheClient.service.ts
+++ b/src/modules/cacheClient/cacheClient.service.ts
@@ -76,7 +76,7 @@ export class CacheClient implements ICacheClient {
     }
 
     /**
-     * @description Interceptor of authentication errors. Schedules failed request and starts authentication
+     * @description Interceptor of authentication errors. Queues failed requests and starts authentication process.
      * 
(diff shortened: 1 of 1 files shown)


src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -76,9 +76,11 @@ export class CacheClient implements ICacheClient {
     }
 
     /**
-     * @description At the time hanldes only authentication errors. Schedules failed requests and starts authentication
-     * @param error
-     * @returns
+     * @description Interceptor of authentication errors. Schedules failed request and starts authentication
```
jrhender, line 79: I think "Queue" describes it better. What do you think?
```suggestion
     * @description Interceptor of authentication errors. Queues failed requests and starts authentication process.
```

jrhender (approved):
Minor suggestion but it looks good to me

src/modules/cacheClient/cacheClient.service.ts
JGiter: :+1: `Queues` is more appropriate term for this

PR #385, opened 2021-12-20
feat(claims): rename fields to requestorFields for better readibility

Passerino:
fields renamed to requestorFields. They will coexist for now for backward compatibility.

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { utils, Wallet } from "ethers";
 import jsonwebtoken from "jsonwebtoken";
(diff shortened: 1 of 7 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -149,14 +151,18 @@ export class ClaimsService {
         claim: {
             claimType: string;
             claimTypeVersion: number;
-            fields: { key: string; value: string | number }[];
+            fields?: { key: string; value: string | number }[];
```
Harasz, line 154: What is the difference between `requestorFields` and `fields` than?

Passerino: fields are left here temporarily, they are supposed to be removed after migration to requestor fields.

Harasz (approved):

PR #388, opened 2021-12-24
feature/ICL-118 remove authentication guide

JGiter:


Harasz (commented):
@JGiter Do we have example for connecting to cache server and DID registry?

Harasz (approved):

PR #390, opened 2021-12-28
feat(cache-client): update test login endpoint

Harasz:


src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -319,8 +319,10 @@ export class CacheClient implements ICacheClient {
      */
     private async isAuthenticated(): Promise<boolean> {
         try {
-            await this.httpClient.get<Asset[]>(`${TEST_LOGIN_ENDPOINT}${this._signerService.did}`);
-            return true;
+            const { data } = await this.httpClient.get<{
```
JGiter, line 322: I guess it should return user for whom access token was issued

JGiter (changes requested):

src/modules/cacheClient/cacheClient.service.ts
jrhender: Hmm, IMO, a name that starts with "is" communicates that the return value is a boolean. Personally, I think the currently implementation is good.

src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -319,8 +319,10 @@ export class CacheClient implements ICacheClient {
...
+                user: string | null;
+            }>(`${TEST_LOGIN_ENDPOINT}`);
+            return data.user === this._signerService.did;
```
jrhender, line 325: In the unlikely event that `this._signerService.did` was `null`, I wouldn't want `isAuthenticated` to return `true`
```suggestion
            return data.user ? data.user === this._signerService.did : false;
```

jrhender (approved):
A small suggestion, but other than that I think it looks good

PR #431, opened 2022-01-14
feat(claims): enable on chain enrolment without previous request

Passerino:
there is on chain an off chain erolment possibility with off chain as default one.

### DIFF
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -20,8 +20,8 @@ import {
 } from '../src';
 import { replenish, root, rpcUrl, setupENS } from './utils/setup_contracts';
 import { ClaimManager__factory } from '../ethers/factories/ClaimManager__factory';
-import { ClaimManager } from '../ethers/ClaimManager';
 import { ProofVerifier } from '@ew-did-registry/claims';
+import { ClaimManager } from '../ethers/ClaimManager';
 
 const { namehash } = utils;
 
@@ -220,6 +220,82 @@ describe('Enrollment claim tests', () => {
       return issuedClaim;
     }
 
+    async function enrolAndIssueWithoutRequest(
+      requestSigner: Required<ethers.Signer>,
+      issueSigner: Required<ethers.Signer>,
+      {
+        subjectDID,
+        claimType,
+        registrationTypes = [RegistrationTypes.OnChain],
+        issuerFields,
+      }: {
+        subjectDID: string;
+        claimType: string;
+        registrationTypes?: RegistrationTypes[];
+        publishOnChain?: boolean;
+        issuerFields?: { key: string; value: string | number }[];
+      }
+    ) {
+      await signerService.connect(issueSigner, ProviderType.PrivateKey);
+      const requesterDID = subjectDID;
+      const issuerDID = signerService.did;
+      const requestorFields = [{ key: 'temperature', value: 36 }];
+
+      await claimsService.issueClaim({
+        subject: subjectDID,
+        registrationTypes,
+        claim: {
+          claimType,
+          claimTypeVersion: version,
+          issuerFields: issuerFields || [],
+        },
+      });
+
+      const [did, issuedClaim] = <[string, Required<IClaimIssuance>]>(
+        mockIssueClaim.mock.calls.pop()
+      );
+
+      expect(did).toBe(issuerDID);
+
+      const { issuedToken, requester, claimIssuer, onChainProof, acceptedBy } = issuedClaim;
+
+      if (registrationTypes.includes(RegistrationTypes.OffChain)) {
+        expect(issuedToken).toBeDefined();
+
+        const { claimData, signer, did } = (await didRegistry.decodeJWTToken({
+          token: issuedToken,
+        })) as { [key: string]: string };
+
+        expect(claimData).toEqual({
+          claimType,
+          claimTypeVersion: version,
+          issuerFields,
+          requestorFields,
+        });
+
+        expect(signer).toBe(issuerDID);
+        expect(did).toBe(requesterDID);
+      }
+      if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+        expect(onChainProof).toHaveLength(132);
+        await signerService.connect(requestSigner, ProviderType.PrivateKey);
+        const mockedClaim = {
+          claimType,
+          isApproved: true,
+          onChainProof,
+          claimTypeVersion: version,
+          acceptedBy: issuerDID,
+        };
+        mockGetClaimsBySubject.mockReset().mockImplementationOnce(() => [mockedClaim]);
+        await claimsService.publishPublicClaim({ claimType, registrationTypes });
+      }
+      expect(requester).toEqual(subjectDID);
+      expect(claimIssuer).toEqual([issuerDID]);
+      expect(acceptedBy).toBe(issuerDID);
+
+      return issuedClaim;
+    }
+
     test('enrollment by issuer of type DID', async () => {
       const requester = rootOwner;
       const issuer = staticIssuer;
@@ -352,6 +428,17 @@ describe('Enrollment claim tests', () => {
         expect(await claimsService.hasOnChainRole(rootOwnerDID, claimType, version)).toBe(true);
       });
 
+      test('should be able to issue and publish onchain without request', async () => {
+        mockGetClaimsBySubject.mockImplementationOnce(() => [role1Claim]); // to verify requesting
+
+        await enrolAndIssueWithoutRequest(rootOwner, staticIssuer, {
+          subjectDID: rootOwnerDID,
+          claimType,
+          registrationTypes,
+        });
+        expect(await claimsService.hasOnChainRole(rootOwnerDID, claimType, version)).toBe(true);
+      });
+
       test('should be able to issue without publishing onchain', async () => {
         mockGetClaimsBySubject.mockImplementationOnce(() => [role1Claim]);
 
@@ -413,10 +500,10 @@ describe('Enrollment claim tests', () => {
 
   describe('Selfsigned claim tests', () => {
     test('Selfsigned claim should be verified', async () => {
-      const claimUrl = await claimsService.createSelfSignedClaim({
+      const claimUrl = (await claimsService.createSelfSignedClaim({
         data: { claimType: roleName1 },
         subject: rootOwnerDID,
-      });
+      })) as string;
       const claim = await didRegistry.ipfsStore.get(claimUrl);
 
       const document = await didRegistry.getDidDocument();
@@ -434,10 +521,10 @@ describe('Enrollment claim tests', () => {
         { document: { id: assetDID }, id: assetDID },
       ]);
       const claimType = 'test claim';
-      const claimUrl = await claimsService.createSelfSignedClaim({
+      const claimUrl = (await claimsService.createSelfSignedClaim({
         data: { claimType },
         subject: assetDID,
-      });
+      })) as string;
       const claim = await didRegistry.ipfsStore.get(claimUrl);
 
       const ownerDoc = await didRegistry.getDidDocument({ did: rootOwnerDID, includeClaims: true });
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -221,7 +221,6 @@ describe('Enrollment claim tests', () => {
     }
 
     async function enrolAndIssueWithoutRequest(
-      requestSigner: Required<ethers.Signer>,
       issueSigner: Required<ethers.Signer>,
       {
         subjectDID,
@@ -278,16 +277,17 @@ describe('Enrollment claim tests', () => {
       }
       if (registrationTypes.includes(RegistrationTypes.OnChain)) {
         expect(onChainProof).toHaveLength(132);
-        await signerService.connect(requestSigner, ProviderType.PrivateKey);
+        
         const mockedClaim = {
           claimType,
           isApproved: true,
           onChainProof,
           claimTypeVersion: version,
           acceptedBy: issuerDID,
+          subject: requesterDID,
         };
         mockGetClaimsBySubject.mockReset().mockImplementationOnce(() => [mockedClaim]);
-        await claimsService.publishPublicClaim({ claimType, registrationTypes });
+        
       }
       expect(requester).toEqual(subjectDID);
       expect(claimIssuer).toEqual([issuerDID]);
@@ -397,8 +397,8 @@ describe('Enrollment claim tests', () => {
         returnSteps: false,
       });
 
-      mockGetClaimsBySubject.mockImplementationOnce(() => []);
-      mockGetClaimsBySubject.mockImplementationOnce(() => []);
+      mockGetClaimsBySubject.mockImplementation(() => []);
+      //mockGetClaimsBySubject.mockImplementationOnce(() => []);
 
       return expect(
         enrolAndIssue(rootOwner, staticIssuer, {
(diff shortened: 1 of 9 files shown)


e2e/claims.service.e2e.ts
```diff
@@ -220,6 +220,82 @@ describe('Enrollment claim tests', () => {
       return issuedClaim;
     }
 
+    async function enrolAndIssueWithoutRequest(
+      requestSigner: Required<ethers.Signer>,
```
JGiter, line 224: I think it is a bit confusing to use `requestSigner` and `requesterDID` in `enrolAndIssueWithoutRequest`. Publishing would be better to be part of the test instead of this method

src/modules/claims/claims.service.ts
```diff
@@ -269,20 +269,50 @@ export class ClaimsService {
   async registerOnchain(
     claim: Pick<Claim, 'token' | 'subjectAgreement' | 'onChainProof' | 'acceptedBy'>
   ) {
-    if (!readyToBeRegisteredOnchain(claim)) {
+    if (!readyToBeRegisteredOnchain(claim) || !claim.token) {
```
JGiter, line 272: I don't think this is necessary as token already required by `readyToBeRegisteredOnchain`

src/modules/claims/claims.service.ts
```diff
@@ -269,20 +269,50 @@ export class ClaimsService {
...
+      subject: sub,
+    });
+  }
+
+  /**
+   * @description Registers issued onchain claim with Claim manager
+   *
+   * @param claimId - id of signed onchain claim
+   */
+  private async performRegisterOnchain(
```
JGiter, line 294: I am personally don't see benefits of moving this out of `registerOnChain`. Do you think `performRegisterOnChain` can be used as not part of `registerOnChain`?

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
     return v4();
   }
 
+  validatePublishPublicClaimRequest(
```
JGiter, line 416: Docs would be good

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
...
+    registrationTypes: RegistrationTypes[],
+    token?: string,
+    claimType?: string
+  ) {
+    if (registrationTypes.includes(RegistrationTypes.OnChain) && !claimType) {
+      throw new Error('claimType required for on-chain registration');
+    }
+    if (registrationTypes.includes(RegistrationTypes.OffChain) && !token) {
```
JGiter, line 424: `token` isn't needed to publish not requested claim

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
...
-    const { iss, claimData } = payload;
-    let sub = payload.sub;
-    // Initialy subject was ignored because it was requester
-    if (!sub || sub.length === 0 || !isValidDID(sub)) {
-      sub = this._signerService.did;
-    }
-    const verifiedDid = await this._didRegistry.verifyPublicClaim(token, iss);
-    if (!verifiedDid || !compareDID(verifiedDid, iss)) {
-      throw new Error('Incorrect signature');
+  async publishPublicClaim({
```
JGiter, line 436: Documentation needs to be adjusted to new method signature

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
...
+    token,
+    claimType,
```
JGiter, line 438: I think that all claim data should be part of the claim

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
     return v4();
   }
 
+  validatePublishPublicClaimRequest(
+    registrationTypes: RegistrationTypes[],
+    token?: string,
+    claimType?: string
+  ) {
+    if (registrationTypes.includes(RegistrationTypes.OnChain) && !claimType) {
+      throw new Error('claimType required for on-chain registration');
```
JGiter, line 422: Please add new error types to `ERROR_MESSAGES` enum

src/modules/claims/claims.service.ts
```diff
@@ -372,46 +413,96 @@ export class ClaimsService {
...
+    registrationTypes = [RegistrationTypes.OffChain],
+  }: {
+    token?: string;
+    claimType?: string;
+    registrationTypes?: RegistrationTypes[];
+  }) {
+    this.validatePublishPublicClaimRequest(registrationTypes, token, claimType);
+    let url: string | undefined = undefined;
+    if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+      const claims = await this.getClaimsBySubject({
```
JGiter, line 448: This would be more structured if `if registrationTypes.includes(RegistrationTypes.OnChain) {` was followed by `registerOnChain({claim})`. Issuance of claim can be checked in `validatePublishPublicClaimRequest`

JGiter (changes requested):

src/modules/claims/claims.service.ts
Passerino: We not always posses token:
in `request -> issue -> publish` scenario we got token (`registerOnChain`)
in `issue -> publish` scenario we don't have token, only `claimType` and `claimTypeVersion` (`performRegisterOnChain`)

I didn't want to amend registerOnchain as it's public, that's why I overloaded it.  The other option would be to accept `claimType` and `claimTypeVersion` instead of token in all cases. Let me know what you think.

src/modules/claims/claims.service.ts
Passerino: sorted

src/modules/claims/claims.service.ts
Passerino: sorted - we will decide on breaking changes to API later on  ( registerOnchain)

src/modules/claims/claims.service.ts
Passerino: needs to be discussed further

src/modules/claims/claims.service.ts
Passerino: +1

src/modules/claims/claims.service.ts
Passerino: +1

e2e/claims.service.e2e.ts
```diff
@@ -321,8 +397,8 @@ describe('Enrollment claim tests', () => {
         returnSteps: false,
       });
 
-      mockGetClaimsBySubject.mockImplementationOnce(() => []);
-      mockGetClaimsBySubject.mockImplementationOnce(() => []);
+      mockGetClaimsBySubject.mockImplementation(() => []);
```
Harasz, line 400: Are we making sure to restore this mocked value since we don't mock it once?

src/modules/claims/claims.service.ts
```diff
@@ -261,28 +262,61 @@ export class ClaimsService {
...
+    acceptedBy?: string;
+    subject?: string;
+  }) {
+    // backward compatibility with token
+    claim = { ...claim, ...this.getClaimTypeFromToken(claim.token) };
+
+    if (
+      !claim.subjectAgreement &&
+      claim.subject === this._signerService.did &&
+      claim.claimType &&
```
JGiter, line 302: We can ensure that `claimType`, `claimTypeVersion` and `subject` are set at this point like this
```
if (!claimType || !...) {
   ({claimType, ...} = this.getClaimTypeFromToken(claim.token));
}
```

src/modules/claims/claims.service.ts
```diff
@@ -261,28 +262,61 @@ export class ClaimsService {
...
-    claim: Pick<Claim, 'token' | 'subjectAgreement' | 'onChainProof' | 'acceptedBy'>
-  ) {
-    if (!readyToBeRegisteredOnchain(claim)) {
-      throw new Error(ERROR_MESSAGES.CLAIM_WAS_NOT_ISSUED);
+  async registerOnchain(claim: {
+    claimType?: string;
+    claimTypeVersion?: string;
+    token?: string;
+    subjectAgreement?: string;
+    onChainProof?: string;
```
JGiter, line 292: `onChainProof` is required for onchain registration

e2e/claims.service.e2e.ts
Passerino: I fixed tests to mockImplementationOnce everywhere

JGiter (changes requested):

e2e/claims.service.e2e.ts
```diff
@@ -434,10 +525,10 @@ describe('Enrollment claim tests', () => {
         { document: { id: assetDID }, id: assetDID },
       ]);
       const claimType = 'test claim';
-      const claimUrl = await claimsService.createSelfSignedClaim({
+      const claimUrl = (await claimsService.createSelfSignedClaim({
         data: { claimType },
         subject: assetDID,
-      });
+      })) as string;
```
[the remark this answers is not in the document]
Passerino: there was a call to `publishPublicClaim` which returned url only in `offChain` registration. In this case though , url is always returned so I reflected types faster.

src/modules/claims/claims.service.ts
```diff
@@ -231,6 +230,8 @@ export class ClaimsService {
       requester,
       claimIssuer: [this._signerService.did],
       acceptedBy: this._signerService.did,
+      claimType: claimData.claimType,
```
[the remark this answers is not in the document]
Passerino: There is validation in cache server against these 2 props. They were taken from token, but it's optional now.

src/modules/claims/claims.service.ts
```diff
@@ -261,28 +262,61 @@ export class ClaimsService {
     return this._cacheClient.issueClaim(this._signerService.did, message);
   }
 
+  /**
+   *
+   * @param token optional token containing claimType, version and subject
+   * @returns claim params obtained from token
+   */
+  private getClaimTypeFromToken(token?: string) {
```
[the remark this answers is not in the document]
Passerino: token is not ment to be required, user can registerOnchain with claim data lacking of token.

src/modules/claims/claims.service.ts
Passerino: yes , onChainProof and acceptedBy changed to required.

src/modules/claims/claims.service.ts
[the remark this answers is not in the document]
JGiter: Hmm, this is strange. Why would cache server to get claim params again? Could you please point where this piece of code. I've found token decoding only when requesting claim

src/modules/claims/claims.service.ts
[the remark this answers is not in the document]
JGiter: `token` is optional because this method is called when registering with and without request. I think that this method should be called only when registering with request

src/modules/claims/claims.service.ts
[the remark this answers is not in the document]
Passerino: I added it to always provide cache server with these values , but as they are not required on issueClaimRequest , I stripped the out as you suggested.

src/modules/claims/claims.service.ts
Passerino: it won't be sure since claim.token is optional. That's why I use token if it exists and then I validate against all required params existance.

JGiter: if (!claimType || !claimTypeVersion || !subject) and !token then error should be thrown

Passerino: Existance of 'claimType', 'claimTypeVersion', 'subject', 'onChainProof', 'acceptedBy', 'subjectAgreement' is validated later on. Before then we are trying to fill the gaps : 'claimType', 'claimTypeVersion', 'subject' from token if possible and 'subjectAgreement' if not provided. 
Then we validate whole request.

src/errors/ErrorMessages.ts
```diff
@@ -19,6 +19,9 @@ export enum ERROR_MESSAGES {
   ERROR_IN_AZURE_PROVIDER = 'Error in Azure Provider',
   JWT_ALGORITHM_NOT_SUPPORTED = 'Jwt algorithm not supported',
   CLAIM_WAS_NOT_ISSUED = 'Claim was not issued',
+  CLAIM_NOT_FOUND = 'Claim not found',
```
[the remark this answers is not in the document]
Passerino: +1

e2e/claims.service.e2e.ts
```diff
@@ -352,6 +428,19 @@ describe('Enrollment claim tests', () => {
         expect(await claimsService.hasOnChainRole(rootOwnerDID, claimType, version)).toBe(true);
       });
 
+      test('should be able to issue and publish onchain without request', async () => {
+        mockGetClaimsBySubject.mockImplementationOnce(() => [role1Claim]); // to verify requesting
```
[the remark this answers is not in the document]
Passerino: +1

e2e/claims.service.e2e.ts
```diff
@@ -220,6 +220,81 @@ describe('Enrollment claim tests', () => {
       return issuedClaim;
     }
 
+    async function enrolAndIssueWithoutRequest(
```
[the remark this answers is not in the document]
Passerino: +1

e2e/claims.service.e2e.ts
```diff
@@ -220,6 +220,81 @@ describe('Enrollment claim tests', () => {
...
+
+        const mockedClaim = {
+          claimType,
+          isApproved: true,
+          onChainProof,
+          claimTypeVersion: version,
+          acceptedBy: issuerDID,
+          subject: requesterDID,
+        };
+        mockGetClaimsBySubject.mockReset().mockImplementationOnce(() => [mockedClaim]);
```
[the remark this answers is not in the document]
Passerino: +1

src/modules/claims/claims.service.ts
```diff
@@ -373,45 +415,95 @@ export class ClaimsService {
...
-      iss: string;
-      sub: string;
-      claimData: ClaimData;
-    };
-    const { iss, claimData } = payload;
-    let sub = payload.sub;
-    // Initialy subject was ignored because it was requester
-    if (!sub || sub.length === 0 || !isValidDID(sub)) {
-      sub = this._signerService.did;
+  private validatePublishPublicClaimRequest(
```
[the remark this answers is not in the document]
Passerino: url = await this._didRegistry.ipfsStore.save(token); 
we need it in off chain, at least according to code.

src/modules/claims/claims.service.ts
```diff
@@ -373,45 +415,95 @@ export class ClaimsService {
...
+   *
+   */
+  async publishPublicClaim({
+    token, // backward compatibility
+    registrationTypes = [RegistrationTypes.OffChain],
+    claim,
+  }: {
+    token?: string;
+    registrationTypes?: RegistrationTypes[];
+    claim: { token?: string; claimType?: string };
```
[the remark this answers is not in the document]
Passerino: place of token is new (migrated from {token} to {claim:{token}} , but I don't know how to make token fully deprecated as it's necessary in off chain journey, isn't it ?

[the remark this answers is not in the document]
JGiter: Ah, indeed. did-registry still asks token for publishing. Perhaps will change this in separate task

JGiter (approved):
Thanks @Passerino . Looks good

PR #432, opened 2022-01-14
bug/ICS-163 asset document not updated

JGiter:
https://energyweb.atlassian.net/browse/ICS-163

Fix isn't properly tested because current test setup doesn't include cache server. Still added simple test for local testing

### DIFF
--- a/e2e/cacheClient.e2e.ts
+++ b/e2e/cacheClient.e2e.ts
@@ -1,0 +1,36 @@
+import { initWithPrivateKeySigner, setCacheConfig } from "../src";
+
+const rpcUrl = "https://volta-rpc.energyweb.org";
+const privateKey = "";
+
+describe.skip("Cache client", () => {
+    test("Asset owner should be able to update asset document", async () => {
+        setCacheConfig(73799, {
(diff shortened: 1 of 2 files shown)


e2e/cacheClient.e2e.ts
```diff
@@ -0,0 +1,36 @@
+import { initWithPrivateKeySigner, setCacheConfig } from "../src";
+
+const rpcUrl = "https://volta-rpc.energyweb.org";
+const privateKey = "";
+
+describe.skip("Cache client", () => {
```
Harasz, line 6: Why `skip`?

JGiter: This test can't be run during CI, because cache server is not included in test setup

Harasz: We could consider using docker here with ICS image. But due to short deadline for Origin team we can postpone this and create task for later

Harasz (approved):

e2e/cacheClient.e2e.ts
JGiter: Yes, this is least what can be done now

PR #433, opened 2022-01-14
fix: setup nats connection options

hejkerooo:


src/modules/messaging/messaging.service.ts
```diff
@@ -1,8 +1,8 @@
-import { Codec, connect, JSONCodec, NatsConnection, Subscription } from 'nats.ws';
-import { getMessagingConfig } from '../../config/messaging.config';
-import { IMessage, MessagingMethod } from './messaging.types';
-import { SignerService } from '../signer/signer.service';
-import { executionEnvironment, ExecutionEnvironment } from '../../utils/detectEnvironment';
```
Harasz, line 5: And why is that?

Harasz (changes requested):

src/modules/messaging/messaging.service.ts
```diff
@@ -37,6 +37,10 @@ export class MessagingService {
             servers: natsServerUrl.replace('http', 'ws'),
             timeout,
             pingInterval: 50 * 1000,
+            reconnect: true,
```
JGiter, line 40: it is true by default

src/modules/messaging/messaging.service.ts
hejkerooo: fixed

Harasz (approved):

JGiter (approved):

PR #438, opened 2022-01-26
fix(claim): fix issuing a claim with on-chain and off-chain request

Harasz:
Replacing the order of role issuing help prevent the error 4001 when logged in using MetaMask.


jrhender (approved):
Doesn't make any sense but can't hurt 😄

Harasz:
As I mentioned on slack putting delay between signatures worked, here we have natural delay from transaction (sending transaction take a couple of ms) so that's why this solution is working. However, I still don't know why delay is solving a problem :/

PR #442, opened 2022-01-31
task/ICL-209 WalletConnect signature is not verifiable

JGiter:


src/modules/signer/signer.service.ts
```diff
@@ -51,8 +51,10 @@ export class SignerService {
     // web app is responsible for clearing of isEthSigner on logout
     if (executionEnvironment() === ExecutionEnvironment.BROWSER) {
       const isEthSigner = localStorage.getItem(IS_ETH_SIGNER);
-      if (isEthSigner) {
-        this._isEthSigner = Boolean(isEthSigner);
+      if (isEthSigner === 'true') {
+        this._isEthSigner = true;
```
Harasz, line 55:
```suggestion
        this._isEthSigner = isEthSigner === 'true';
```

Harasz (changes requested):

src/modules/signer/signer.service.ts
Harasz: We can simplify it to one line

JGiter: I am not sure this will work. We need sign differently when `_isEthSigner` is true, false and undefined. But your change means that _isEthSigner will set to false, when isEthSigner is undefined

Harasz: But we want to have as a `true` value when it's undefined?

Harasz: what about `this._isEthSigner =  ['true', undefined].includes(isEthSigner)`? I dont like current approach but I can't find a better way

JGiter: We need strict correspondence between `isEthSigner` and `_isEthSigner` except conversion `true` -> true, `false` -> false. Your last proposal results in either `true` or `false` (as return type of `includes`) and `undefined` will be lost

Harasz: Why do you want to have `this._isEthSigner` as `undefined`? It should be just `true` or `false` as type declaration says `private _isEthSigner: boolean;`

JGiter: `undefined` means that signer type has not yet been determined and therefore the first signature should be requested

Harasz: Okay then we should change type to `private _isEthSigner?: boolean;` and we can proceed with this PR

Harasz (approved):

src/modules/signer/signer.service.ts
JGiter: I prefer to think about `_isEthSigner: boolean` as type of member of initialized `CacheServerClient`. Having `?` in type might make think that `_isEthSigner` of fully initialized `CacheServerClient` is underfined

Harasz: Hmm, but it's not set in the `init` method in the end, which can be confusing. Maybe there is a better way to handle this. But for now as a hotfix is okay for me

PR #444, opened 2022-01-31
Task/icl 211 update node npm

Harasz:


jrhender (approved):
Nice! What was the solution to the errors you were having in the end?

Harasz:
I have to install `@ethersproject/abstract-provider`, `@ethersproject/abstract-signer` and `@ethersproject/providers` as our dependency and the same version as it has in `ethers`. Also I have to update provider type in EKC signer.

PR #449, opened 2022-02-01
bug/ICL-223 v4.1 breaking changes were not added to changelog

JGiter:
https://energyweb.atlassian.net/browse/ICL-223

Changes in question https://github.com/energywebfoundation/iam-client-lib/pull/431

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -20,8 +20,15 @@
 ### Features
 
 * **claims:** enable on chain enrolment without previous request ([b80446d](https://github.com/energywebfoundation/iam-client-lib/commit/b80446da278d7435d98723ef8f1bede5b4aa22d2))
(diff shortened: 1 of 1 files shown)


JGiter:
@jrhender It appears that not only Changelog was not updated, but instead of major the minor version was bumped. I could commit this PR as breaking change to fix that. What do you think?

Harasz:
@JGiter I think we should revert #431 and commit changes again but we braking change mark.

jrhender:
> @JGiter I think we should revert #431 and commit changes again but we braking change mark.

I think that probably this is the best way, generally speaking. But as we've already merged a "breaking change" PR (https://github.com/energywebfoundation/iam-client-lib/pull/450), I don't think it matters much. Which ever you guys prefer.

jrhender (approved):

PR #456, opened 2022-02-09
feat(did): getters and setters to DID publicKey, delegate, claims

Passerino:
getters for claims, publicKey and delegate
setters for publicKey and delegate
fix to nats build script


src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -102,6 +104,31 @@ export class DidRegistry {
     };
   }
 
+  /**
+   * @description get claims from User's DID document
+   * @returns list of claims
+   */
+  async getDidClaims({ did = this._signerService.did }: { did?: string } | undefined = {}) {
```
JGiter, line 111: Since this actually returns service endpoints I suggest rename it to `readServices`. We can also add method to return actual offchain claims. I think for this we can extend `ClaimsService.getClaims...` with `registrationTypes` parameter. From my perspective ther is no need to wrap `did` in object

src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -199,6 +226,58 @@ export class DidRegistry {
     }
   }
 
+  /**
+   * @description updates public key of currently signed did document
```
JGiter, line 230:
```suggestion
   * @description Adds public key to the document of controlled `did`
```

src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -199,6 +226,58 @@ export class DidRegistry {
...
+   * @returns true if document is updated successfuly
+   */
+  async updateSignedDidPublicKey({
```
JGiter, line 233: This could be more general if could add key to document of any controlled identity

JGiter (changes requested):

src/modules/didRegistry/didRegistry.service.ts
jrhender: My thoughts are:
- I definitely agree wth the rename to `readServices`.
- I don't think we should add a method to return actual offchain claims at this point as this will not be possible once we move to a presentation flow.

src/modules/didRegistry/didRegistry.service.ts
jrhender: Great point 👍

jrhender (changes requested):
I agree with Dmitry's position, this probably needs some changes. I added a suggestion 😄

src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -102,6 +104,31 @@ export class DidRegistry {
...
+   */
+  async getDidPublicKeys({ did = this._signerService.did }): Promise<IPublicKey[]> {
+    return (await this.getDidDocument({ did: did, includeClaims: false })).publicKey;
+  }
+
+  /**
+   * @description get public keys from DID's document
+   * @returns list of DID's delegates
+   */
+  async getDidDelegetes({ did = this._did }): Promise<string[] | undefined> {
```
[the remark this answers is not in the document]
Passerino: +1

src/modules/didRegistry/didRegistry.service.ts
Passerino: +1

JGiter (approved):

PR #459, opened 2022-02-11
task/ICS-178 migrate from PublicResolver

JGiter:
https://energyweb.atlassian.net/browse/ICS-178

- all definitions on `PublicResolver` and `RoleDefinitionResolver` are migrated to `RoleDefinitionResolver_v2`
- updated legacy issuers DID's
- updated legacy role versions

In order to not re-migrate script will be executed when iam-contracts (or iam-roles) will be changed to create new domains on RoleDefinitionResolverv2


src/utils/domainTraverse.ts
```diff
@@ -0,0 +1,166 @@
...
+          did: did?.map((d) =>
+            isValidDID(d) ? d : `did:${Methods.Erc1056}:${chainName}:${d.split(':').slice(-1)}`
+          ),
+        },
+        version: parseInt(version.toString(), 10),
+      };
+      const updateTx = transactionFactory.editDomain({
+        domain,
+        domainDefinition: updatedDef,
+      });
```
jrhender, line 56: Instead of just `editDomain`, I think there is some additional work that needs to be done to change ENSRegistry to point to new resolver and set reverse name. See https://github.com/energywebfoundation/iam-client-lib/blob/eae6b169f5c9e960655cd38a6f8d414297bd55d1/src/modules/domains/domains.service.ts#L899

src/utils/updateLegacyIssuers.ts
```diff
@@ -0,0 +1,177 @@
...
+    transactionFactory,
+    chainName,
+    provider,
+    domainHierarchy,
+    domainReader,
+    ensRegistry,
+    ensResolverV2Address,
+  } = await initDomains(signer, chainId);
+
+  let updated = new Array<any>();
```
Harasz, line 38:
```suggestion
  const updated = new Array<any>();
```
We can also an actual definition of the array items and not a `any`.

Harasz (changes requested):

src/utils/updateLegacyIssuers.ts
```diff
@@ -0,0 +1,177 @@
...
+          issuer: {
+            ...def.issuer,
+            did: did?.map((d) =>
+              isValidDID(d) ? d : `did:${Methods.Erc1056}:${chainName}:${d.split(':').slice(-1)}`
+            ),
+          },
+          version: parseInt(version.toString(), 10),
+        };
+        updated.push({ domain, legacyDef: def, updatedDef });
+        updateDomainTx = transactionFactory.newRole({ domain, roleDefinition: updatedDef });
```
jrhender, line 85: @JGiter Is a tx to the ENS Registry also required to update resolver that the domain is using? I don't think that this is currently being covered. https://github.com/energywebfoundation/iam-client-lib/blob/eae6b169f5c9e960655cd38a6f8d414297bd55d1/src/modules/domains/domains.service.ts#L908

jrhender (changes requested):

src/utils/updateLegacyIssuers.ts
jrhender: Oh nvm, sorry, I see that you are doing it already with `await ensRegistry.setResolver(domain, ensResolverV2Address)`

src/utils/updateLegacyIssuers.ts
```diff
@@ -0,0 +1,177 @@
...
+        } = def;
+        const updatedDef: IRoleDefinition = {
+          ...def,
+          issuer: {
+            ...def.issuer,
+            did: did?.map((d) =>
+              isValidDID(d) ? d : `did:${Methods.Erc1056}:${chainName}:${d.split(':').slice(-1)}`
+            ),
+          },
+          version: parseInt(version.toString(), 10),
```
jrhender, line 82: I tested a bit and I think this would work if `version` is something like `1.0.0` which I'm a bit surprised by. But I guess if it works 👍 !

src/config/chain.config.ts
```diff
@@ -15,6 +15,7 @@ import {
   EWC_IDENTITY_MANAGER_ADDRESS,
   EWC_PUBLIC_RESOLVER_ADDRESS,
   EWC_RESOLVER_V1_ADDRESS,
+  VOLTA_RESOLVER_V2_ADDRESS,
```
jrhender, line 18: I didn't realize that we weren't using the `V2` Resolver in ICL 🤯 ! @nichonien do you think it would be a good idea if we started using it so that we're better placed to support revocation?

jrhender (approved):
This looks good to me! I think we should think about changing the resolver contract that ICL uses for new domains to the `V2` address before we run this script though so that we can be sure that all role definitions will be on the `V2` Resolver going forward. What do you think?

src/utils/updateLegacyIssuers.ts
Harasz: I was also surprised :smile:

src/config/chain.config.ts
nichonien: @jrhender Yeah, definitely. We need it for revocation.

src/utils/updateLegacyIssuers.ts
JGiter: Yeah, this let/const a bit confusing in js. I removed `any` though to not overcomplicate, because `updated` is only used for logging

src/utils/updateLegacyIssuers.ts
JGiter: It just find first dot and strips it along with rest

Harasz (approved):

PR #472, opened 2022-02-21
build: add WC patches to build

Harasz:
I am not sure if this gone work, but it's hard to test it locally


JGiter (commented):
Have you tried --save-bundle option? https://docs.npmjs.com/cli/v8/using-npm/config#save-bundle

Harasz:
I don't think that is the best case for bundled dependencies. https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundleddependencies

JGiter:
> I don't think that is the best case for bundled dependencies. https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundleddependencies

Right, it will add 200Kb to 1.15Mb of current iam-client-lib size. So if I understand problem correctly we need to specify correct path to directory where patch files are located. If this so, then we can replace `patch-package` command in `prepare` script with script which will locate `iam-client-lib/patches` directory and set correct `--patch-dir` in `patch-package`. What is your opinion on this?

Harasz:
The problem is that it will break our local package patching. I want to publish this version and test live against SB to see which hook will trigger patch-package after installation.

JGiter (approved):
Ok. Let's give it a try

PR #473, opened 2022-02-21
bug/IS-32 Metamask transaction revert reason is not specified

JGiter:
https://energyweb.atlassian.net/browse/IS-32


jrhender:
@JGiter should we close this? From internal discussions, it seems this was related to chain infrastructure and not to the client code

PR #477, opened 2022-02-22
build: add build patches script

Harasz:


scripts/patches-build.js
```diff
@@ -0,0 +1,35 @@
...
+      if (err) reject(err);
+      const newData = data.replaceAll('node_modules', '..');
+      fs.writeFile(`./dist/patches/${file}`, newData, 'utf8', (err) => {
+        if (err) reject(err);
+        console.log(`Updated ${file}`);
+        resolve();
+      });
+    });
+  });
+}
```
jrhender, line 27: @Harasz It looks we're basically writing our own `patch-package` here? Is this a fair assessment?

jrhender: Ah okay, gotcha. Maybe I'm just bad at reading code but I think that maybe a comment can't hurt
```suggestion
/**
 * Updates the patchFile so that patch-package can notice them from node_modules directory
 */
async function updatePatchFile(file) {
  console.log(`Updating ${file}...`);
  return new Promise((resolve, reject) => {
    fs.readFile(`./dist/patches/${file}`, 'utf8', (err, data) => {
      if (err) reject(err);
      const newData = data.replaceAll('node_modules', '..');
      fs.writeFile(`./dist/patches/${file}`, newData, 'utf8', (err) => {
        if (err) reject(err);
        console.log(`Updated ${file}`);
        resolve();
      });
    });
  });
}
```

jrhender (approved):

scripts/patches-build.js
Harasz: maybe I am bad at writing or documenting a code :smile:

PR #478, opened 2022-02-22
task/ICL-251 update resolver to v2

JGiter:
https://energyweb.atlassian.net/browse/ICL-251

- New domain will be defined on RoleDefinitionResolverV2
- Until domains are not migrated they will be read from RoleDefinitionResolverV2 and RoleDefinitionResolver

This should be followed by corresponding update of ssi-hub


src/modules/domains/domains.service.ts
```diff
@@ -75,27 +78,20 @@ export class DomainsService {
...
-        type: ResolverContractType.PublicResolver,
-      });
-    ensResolverAddress &&
-      this._domainDefinitionReader.addKnownResolver({
-        chainId,
-        address: ensResolverAddress,
-        type: ResolverContractType.RoleDefinitionResolver_v1,
+        address: ensResolverV2Address,
+        type: ResolverContractType.RoleDefinitionResolver_v2,
       });
```
jrhender, line 86: I'm not sure we want to remove the `PublicResolver` code yet as we have yet executed the script. I think I would do this in 3 steps:
1. Use V2 Resolver for Volta
2. Update all of the old domain defs on Volta to use the V2 Resolver
3. Remove the PublicResolver code from ICL (as you are doing in this PR)

What are you thoughts on this approach?

src/modules/domains/domains.service.ts
```diff
@@ -75,28 +85,48 @@ export class DomainsService {
       ensRegistryAddress,
       provider,
     });
-    ensPublicResolverAddress &&
+    this._ensRegistryAddress = ensRegistryAddress;
+    if (chainId === EWC_CHAIN_ID) {
```
jrhender, line 89: Hmm, I think we should try to avoid this special conditional for EWC chain. Would the changes from https://github.com/energywebfoundation/iam-contracts/pull/113 help with avoiding this?

Harasz (approved):

src/modules/domains/domains.service.ts
```diff
@@ -78,28 +87,31 @@ export class DomainsService {
       ensRegistryAddress,
       provider,
     });
-    ensPublicResolverAddress &&
-      this._domainDefinitionReader.addKnownResolver({
-        chainId,
-        address: ensPublicResolverAddress,
-        type: ResolverContractType.PublicResolver,
-      });
```
jrhender, line 86: @JGiter Thanks for making the changes. I think we need to support the `PublicResolver` though until we're pretty sure that we've migrated all of the old domains though, right? Then we can remove the `PublicResolver` code?

jrhender (changes requested):

src/modules/domains/domains.service.ts
JGiter: Ok

src/modules/domains/domains.types.ts
```diff
@@ -38,3 +44,18 @@ export interface IApp {
 }
 
 export const NODE_FIELDS_KEY = 'metadata';
+
+export function castToV2(
```
jrhender, line 48:
```suggestion
// TODO: remove once all of the VOLTA roles have been migrated to v2
export function castToV2(
```

jrhender (approved):
Looks good to me! Just one comment for suggested code comment

PR #481, opened 2022-02-23
revert patch-package fix

JGiter:


Harasz:
It doesn't seems like it reverting. I will psuh my PR with fixes

Harasz:
https://github.com/energywebfoundation/iam-client-lib/pull/482

PR #484, opened 2022-02-24
fix: apply patches by git command

Harasz:


package.json
```diff
@@ -30,6 +30,7 @@
     "build:typechain:publicresolver": "typechain --target ethers-v5 --out-dir ethers \"./node_modules/@energyweb/iam-contracts/dist/contracts/PublicResolver.json\"",
     "build:typechain:staking-pool": "typechain --target ethers-v5 --out-dir ethers-staking \"./node_modules/@energyweb/staking-pool/dist/contracts/StakingPool.sol/StakingPool.json\"",
     "prepare": "npm run build && patch-package",
+    "preinstall": "(git apply --ignore-whitespace dist/patches/@walletconnect+ethereum-provider+1.7.1.patch || true) && (git apply --ignore-whitespace dist/patches/@walletconnect+jsonrpc-provider+1.0.0.patch || true)",
```
jrhender, line 33: Shouldn't it be in `postinstall`? Because the `node_modules` won't be there in `preinstall`?

jrhender (commented):
- Are the `patch-package` patches in the correct format to be understood by `git apply`?
- This could be a separate PR, but would it be possible to have a script that loops over the `dist/patches` dir and applies the patches instead of hardcoding?

package.json
Harasz: Yeah, good catch I will fix that

Harasz:
@jrhender According to the `patch-package` documentation it is supported https://github.com/ds300/patch-package#notes.
If this approach will work then I can make a script for that

jrhender (approved):

PR #485, opened 2022-02-24
chore: add pull request checklist

Harasz:
My proposal for pull request check list. Let me know what you think and feel free to propose your points.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+### Description
+
+<!-- PUT YOUR TASK DESCRIPTION HERE -->
+
+### Contributor checklist
+
+- [ ] Breaking changes - check for any existing interfaces changes that are not backward compatible, removed method etc.
+- [ ] Documentation - document your code, add comments for method, remember to check if auto generated docs were updated.
+- [ ] Tests - add new or updated existed test for changes you made.
+- [ ] Migration guide - add migration guide for every breaking change.
+- [ ] Configuration correctness - check that any configuration changes are correct ex. default URLs, chain ids, smart contract verification on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).


jrhender (commented):
Why in the `.github` directory? Do we want it to be hidden? Maybe it's better directly in the root https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

Harasz:
@jrhender I followed convention from the `ew-did-registry` repo. We can put it anywhere from I can see.

Harasz:
What about putting just the list to the README under contributor section? Bc I don't think we want to store everything in the root such as templates

jrhender:
Hmm, but if we put it in the README, then it would automatically populate in the PR description, right? For that I think it needs to be in a `pull_request_template.md` file?

Harasz:
I meant leave the current file and add just a list to README. So we will have it in template and README

jrhender (approved):
Hmm, I don't want to duplicate the checklist in two spots (README and template). No worries, under `.github` is fine, maybe its better actually

Harasz:
Let see what others thinks on that

dawidgil:
It looks like [Angular team](https://github.com/angular/angular/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and [React team](https://github.com/facebook/react/tree/main/.github) have it inside .github directory. Probably most of libraries have done it this way, so imo let's follow convention

dawidgil (approved):

JGiter (approved):
Looks fine. But IMO `Smart contract addresses` would be named as `Configuration`

Harasz:
@JGiter I like this idea, it would be something like:

- [ ] Configuration - check any configuration that you changes ex. ssi-hub URL, chain id, smart contract addresses that you added/updated on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).


What you think guys? @jrhender @dawidgil

jrhender:
> @JGiter I like this idea, it would be something like:
> 
> * [ ]  Configuration - check any configuration that you changes ex. ssi-hub URL, chain id, smart contract addresses that you added/updated on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).
> 
> What you think guys? @jrhender @dawidgil

I don't really get it. What would the "configuration" check be for? To make sure that the configuration change isn't a breaking change? Or just that it is documented correctly?

Harasz:
To check for example if default smart contract addresses are correct. We have a situation last week that we merged wrong address. To this checking we can include all URL for example SSI Hub, chain ids

jrhender:
Ok, I see. I would suggest then:

 * [ ]  Configuration correctness - check that any configuration changes are correct ex. default URLs, chain ids, smart contract addresses

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+
+<!-- PUT YOUR TASK DESCRIPTION HERE -->
+
+### Contributor checklist
+
+- [ ] Breaking changes - check for any existing interfaces changes that are not backward compatible, removed method etc.
+- [ ] Documentation - document your code, add comments for method, remember to check if auto generated docs were updated.
+- [ ] Tests - add new or updated existed test for changes you made.
+- [ ] Migration guide - add migration guide for every breaking change.
+- [ ] Smart contract addresses - check every smart contract addresses that you updated on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).
```
jrhender, line 11: Is this meant to be about verification? 
```suggestion
- [ ] Smart contract verification - check that any new smart contracts have been verified on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).
```

PR #494, opened 2022-03-01
chore: VC POC

Harasz:


e2e/claims.service.e2e.ts
```diff
@@ -113,7 +114,7 @@ jest.mock('../src/modules/messaging/messaging.service', () => {
   };
 });
 
-describe('Enrollment claim tests', () => {
+describe.skip('Enrollment claim tests', () => {
```
jrhender, line 117: Just a note to remove the `skip` before merging
```suggestion
describe('Enrolment claim tests', () => {
```

e2e/fixtures/externalVC.ts
```diff
@@ -0,0 +1,175 @@
+export const validExemplaryExternalVC = {
```
jrhender, line 1: Minor name change suggestion. To me, "Exemplary" suggests that it is an ideal or very good credential. Is that what you mean?

Were these credentials copied or generated from somewhere? If so, I suggest putting a link to that spot or code.

```suggestion
export const validExampleExternalVC = {
```

src/modules/signer/signer.types.ts
```diff
@@ -33,3 +36,5 @@ export type AccountInfo = {
 
 export const PUBLIC_KEY = 'PublicKey';
 export const IS_ETH_SIGNER = 'isEthSigner';
+
+export type SignerT = Signer & TypedDataSigner;
```
jrhender, line 40: What is the relationship between `Signer` and `TypedDataSigner`? Could we just use `TypedDateSigner`?

What is the `T` meant to mean in `SignerT`?

jrhender (commented):
Great work! Some thoughts/questions

jrhender:
@Harasz What would it take for this PR to be mergeable in your view?

e2e/fixtures/externalVC.ts
Harasz: I copied it from one of the spruce project. And yeah it is just an example, but working one so I think we can think of it like "very good" :smile: I will rename it

src/modules/signer/signer.types.ts
Harasz: I couldn't find a good name for this type, so for now I left it as `SignerT`. Here is the code of the `TypedDataSigner`
```
export interface TypedDataSigner {
    _signTypedData(domain: TypedDataDomain, types: Record<string, Array<TypedDataField>>, value: Record<string, any>): Promise<string>;
}
``` 
And here is a comment for it in ethers lib 
```
// @TODO: This is a temporary measure to preserse backwards compatibility
//        In v6, the method on TypedDataSigner will be added to Signer
```

Harasz:
@jrhender More tests. I am still checking it against Metamask and walletconnect. The next thing is to make it work on both Node and browser.

Harasz:
Here is the first working draft for VC

src/modules/verifiableCredentials/index.ts
```diff
@@ -0,0 +1,27 @@
+export * from './verifiableCredentials.types';
+export * from './verifiableCredentialsBase.service';
+
+import { SignerService } from '../signer';
+import { VerifiableCredentialsServiceBase } from './verifiableCredentialsBase.service';
+
+export const getVerifiableCredentialsService = async (
+  signerService: SignerService
+): Promise<VerifiableCredentialsServiceBase> => {
+  const isNode =
```
JGiter, line 10: you can use https://github.com/energywebfoundation/iam-client-lib/blob/c45a81c36af782cc8d93d3e943e331789e66c571/src/utils/detectEnvironment.ts#L6

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export interface VerifiableCredential<T> {
+  '@context': string | string[];
+  id?: string;
+  issuer: string;
+  type: string[];
+  credentialSubject: T;
+}
+
+export interface Domain {
```
JGiter, line 11: it is already defined in `ethers` https://github.com/ethers-io/ethers.js/blob/dfaa8ee7e6f808dc7f363863243ea650bc629197/packages/abstract-signer/src.ts/index.ts#L25

package.json
```diff
@@ -66,21 +66,24 @@
...
     "@gnosis.pm/safe-apps-provider": "0.8.0",
     "@gnosis.pm/safe-apps-sdk": "4.3.0",
     "@metamask/detect-provider": "^1.2.0",
-    "@walletconnect/ethereum-provider": "^1.7.1",
+    "@walletconnect/ethereum-provider": "1.7.1",
+    "@walletconnect/jsonrpc-provider": "1.0.0",
     "axios": "^0.26.0",
     "base64url": "^3.0.1",
     "concurrently": "^6.1.0",
+    "didkit-wasm": "^0.1.9",
```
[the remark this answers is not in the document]
JGiter: nevermind, it is changed in vp PR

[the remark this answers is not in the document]
Harasz: Here is the reason https://github.com/spruceid/didkit/issues/272. From `@spruceid/didkit-wasm` we are only using `verify` methods

src/modules/verifiable-credentials/eip712.types.ts
```diff
@@ -0,0 +1,92 @@
+export const verifiableCredentialEIP712Types = {
+  EIP712Domain: [],
```
JGiter, line 2: Could you please elaborate what is the purpose of this type and why it is deleted domain types before signing?

Harasz: This type is required by Spruce library, but `ethers` don't want it.

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -0,0 +1,314 @@
...
+    signature: string
+  ) => Promise<string>;
+
+  /**
+   * @param {string} vc - The verifiable credential object in JSON format
+   * @param {string} proof_options - The proof options in JSON format
+   * @description verify given verifiable credential
+   * @returns {Promise<string>} object with results of verification  in JSON format
+   */
+  protected abstract verifyCredential: (
```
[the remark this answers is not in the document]
Harasz: I don't know what you mean. Credential and presentation are designed in the same way. The only difference is for VC we can use DIDKit, but for presentation we can't because DIDKit doesn't provide yet a methods for that. So basically 
`prepareIssuePresentation` and ` completeIssuePresentation` can be removed for now, but when the Spruce team resolve two issues that we reported there will be possible to use `prepareIssuePresentation` and ` completeIssuePresentation`.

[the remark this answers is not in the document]
JGiter: I mean why `verifyCredential:(`, but `verifyPresentation(`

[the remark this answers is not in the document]
Harasz: Oh I see, thanks for noticing that. Will update

src/utils/updateLegacyIssuers.ts
```diff
@@ -5,17 +5,27 @@ import {
   IRoleDefinitionV2,
   ResolverContractType,
 } from '@energyweb/iam-contracts';
```
nichonien, line 7:
```suggestion
} from '@energyweb/credential-governance';
```

src/modules/verifiable-credentials/verifiable-credentials.types.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface EWFCredentialSubject extends ICredentialSubject {
+  role: {
+    namespace: string;
+    version: string;
+  };
+  issuerFields: IssuerFields[];
+  id: string;
+}
+
+export interface VerifiableCredential<T extends ICredentialSubject>
```
nichonien, line 33: @Harasz Do you think, if `EWCredentialSubject` could be used in place of `ICredentialSubject`? I think the prior will suffice our requirements.

Harasz: Definitely, but if we have a `ICredentialSubject` it's more generic.

jrhender:
@Harasz Can we remove the "DO NOT MERGE" label from your perspective? Or is that label still appropriate?

src/modules/verifiable-credentials/verifiable-credentials.types.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface IssuerFields {
+  key: string;
+  value: string | number;
+}
+export interface CreateCredentialParams {
+  subject: string;
+  namespace: string;
+  version: string;
+  issuerFields?: IssuerFields[];
+}
```
jrhender, line 18: I would move these types, as well as `EWFCredentialSubject` to a separate `role-credential.types.ts` file. This can keep `verifiable-credentials.types.ts` to be more generic and organized.

I also think that `CreateCredentialParams` should be `RoleCredentialSubjectParams` as these aren't params for a general credential.

I think we should also just consider using the `EWFCredentialSubject`/`RoleCredentialSubject` directly and removing the separate params interface as they have the same fields essentially. What do you think of this?

```suggestion
export interface IssuerFields {
  key: string;
  value: string | number;
}
/*
 * Parameters required to construct the subject for a role credential
 * https://www.w3.org/TR/vc-data-model/#credential-subject
 */
export interface RoleCredentialSubjectParams {
  /*
   * https://www.w3.org/TR/vc-data-model/#identifiers
   */
  id: string;
  namespace: string;
  version: string;
  issuerFields?: IssuerFields[];
}
```

src/modules/verifiable-credentials/verifiable-credentials.types.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface Credential<T extends ICredentialSubject>
+  extends IJsonLdCredential {
+  credentialSubject: T;
+}
+
+export interface EWFCredentialSubject extends ICredentialSubject {
```
jrhender, line 24: Better to call it `RoleCredentialSubject`. EWF can be about more than just roles 😄 
```suggestion
export interface RoleCredentialSubject extends ICredentialSubject {
```

src/modules/verifiable-credentials/verifiable-credentials.types.ts
```diff
@@ -0,0 +1,97 @@
...
+  role: {
+    namespace: string;
+    version: string;
+  };
+  issuerFields: IssuerFields[];
+  id: string;
```
jrhender, line 30:
```suggestion
  /*
   * https://www.w3.org/TR/vc-data-model/#identifiers
   */
  id: string;
```

jrhender (approved):
Some further suggestions but I think it looks great and we can make further incremental improvements as we go along 😄

src/modules/verifiable-credentials/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  let service: VerifiableCredentialsServiceBase;
+  if (executionEnvironment() === ExecutionEnvironment.NODE) {
+    service = await import('./verifiable-credentials-node.service').then(
+      (module) => module.VerifiableCredentialsServiceNode.create(signerService)
+    );
+  } else {
+    service = await import('./verifiable-credentials-web.service').then(
+      (module) => module.VerifiableCredentialsServiceWeb.create(signerService)
+    );
+  }
```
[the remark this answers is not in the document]
Harasz: No exactly, Spruce has separate library for node and web.

src/modules/verifiable-credentials/verifiable-credentials.types.ts
Harasz: These types will be moved to `ew-did-registry/credential-interface`

src/modules/verifiable-credentials/index.ts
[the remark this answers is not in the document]
jrhender: Ah sorry, I was confused then.
https://github.com/spruceid/ssi/issues/405 is just whether or not we can use the latest version of their libraries then?

src/modules/verifiable-credentials/verifiable-credentials.types.ts
jrhender: @Harasz I don't think we should move types related specifically to role credentials to `ew-did-registry/credential-interface`. I think `ew-did-registry` should remain generic "SSI", not having EW specific details. IMO, it would be better to have these types, as they relate to role credentials, somewhere in [ew-credentials](https://github.com/energywebfoundation/ew-credentials/tree/develop/packages)
Would you agree with this @JGiter @nichonien ?

src/modules/verifiable-credentials/index.ts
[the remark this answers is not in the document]
Harasz: Exactly, right now we are using deprecated library for VC creation. Also it will allow us to use Spruce for VP creation.

src/modules/verifiable-credentials/verifiable-credentials.types.ts
Harasz: Yeah, sorry. I thought they're already in `@ew-did-registry/credential-interface`. I'll move it to separate file then

JGiter: As `IRole` is defined in `iam-client-lib` I think types related to `IRole` also should be defined in `iam-client-lib`

JGiter (approved):

src/modules/verifiable-credentials/verifiable-credentials-node.service.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) => Promise<string>;
+  protected verifyPresentation: (
+    vp: string,
+    proof_options: string
+  ) => Promise<string>;
+
+  constructor(_signerService: SignerService) {
+    super(_signerService);
+
+    this.completeIssueCredential = completeIssueCredential;
```
JGiter, line 48: Just to doublecheck: are you sure that this implements `completeIssueCredential` from base class?

Harasz: Yes, I am

PR #497, opened 2022-03-02
feat(did): update did document params validation

Passerino:
### Description

checking params on updateSignedDidPublicKey, updateSignedDidDelegate functions
checking params on updateDocument conditionally depending on didAttribute value
in case of validation issues - throws Error with the list of invalid or missing properties

### Contributor checklist

- [ v ] Breaking changes - no BC only thing is no more false-positive DID updates. In case there was one, there will be error thrown instead
- [ v ] Documentation - document your code, add comments for method, remember to check if auto generated docs were updated.
- [ v ] Tests - regression tests were used to verify the code
- [ v ] Migration guide - not apply
- [ v ] Configuration correctness - not apply

### DIFF
--- a/src/modules/didRegistry/didRegistry.service.ts
+++ b/src/modules/didRegistry/didRegistry.service.ts
@@ -3,6 +3,7 @@ import { AxiosError } from 'axios';
 import { KeyType } from '@ew-did-registry/keys';
 import { JWT } from '@ew-did-registry/jwt';
 import { ProxyOperator } from '@ew-did-registry/proxyidentity';
+
 import {
   addressOf,
   EwSigner,
@@ -37,8 +38,13 @@ import { ClaimData } from '../didRegistry/did.types';
 import { chainConfigs } from '../../config/chain.config';
 import { AssetsService } from '../assets/assets.service';
 
-const { JsonRpcProvider } = providers;
+import {
+  UpdateServicePoint,
+  UpdateDelegate,
+  UpdatePublicKey,
+} from './didRegistry.validation';
 
+const { JsonRpcProvider } = providers;
 export class DidRegistry {
   private _identityOwner: EwSigner;
   private _operator: Operator;
@@ -89,28 +95,6 @@ export class DidRegistry {
     this.setJWT();
     await this._setDocument();
     this._setClaims();
-  }
-
-  private async getDIDDocFull(did) {
-    if (did === this._signerService.did) {
-      return this._document;
-    } else {
-      const assetDID = (await this._assetsService.getOwnedAssets()).find(
-        (a) => a.document.id === did
-      )?.id;
-      if (!assetDID) {
-        throw new Error(ERROR_MESSAGES.CAN_NOT_UPDATE_NOT_CONTROLLED_DOCUMENT);
-      }
-
-      const { didRegistryAddress: didContractAddress } =
-        chainConfigs()[this._signerService.chainId];
-      const operator = new ProxyOperator(
-        this._identityOwner,
-        { address: didContractAddress },
-        addressOf(assetDID)
-      );
-      return new DIDDocumentFull(did, operator);
-    }
   }
 
   async getDidDocument({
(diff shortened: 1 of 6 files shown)


src/modules/didRegistry/didRegistry.service.ts
```diff
@@ -418,4 +432,56 @@ export class DidRegistry {
       })
     );
   }
+
+  /**
+   * validates update document request
+   */
+  private validDateUpdateDocumentRequest({
```
JGiter, line 439: There is a package that can help with this https://www.npmjs.com/package/ts-interface-checker. Do you think we can use it?

JGiter (commented):
Since document can be updated in did-registry directly does it make sense to put validation logic there?

src/modules/didRegistry/didRegistry.service.ts
Passerino: At first glance it is missing : nested objects validation, enums support and conditions (which are not crucial). I'll give it a try to see if the code improves.

Passerino: it has got enums support, promising ...

Passerino: nested objects too ...

src/modules/didRegistry/didRegistry.validation.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  DIDAttribute,
+  Encoding,
+  PubKeyType,
+} from '@ew-did-registry/did-resolver-interface';
+import { KeyType } from '@ew-did-registry/keys';
+import { createCheckers, iface, enumtype } from 'ts-interface-checker';
+
+const IS_REQ_STRING = 'string';
+export const { UpdateServicePoint, UpdateDelegate, UpdatePublicKey } =
```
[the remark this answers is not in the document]
Passerino: Autogenerated checker would only verify if object is aligned with it's TS definition. It won't validate anything more than type of props. 
As far as I'm concerned, what we need is:
- conditional validation depending on `DIDAttribute` - Claim, PublicKey or Delegate
- confirmation if specific enum is of specific value
- depending on didAttribute some of optional props are in fact required
The only way we could achieve it with autogeneration is to 
- split Claim, PublicKey and Delegate into 3 separate interfaces, but this would require from us a lot of effort and probably would be against the initial architectural decisions. 
- or we could generate 3 interfaces only for the purpose of validation, but this would be more confusing than current solution.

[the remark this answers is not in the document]
JGiter: Make sense

src/modules/didRegistry/didRegistry.validation.ts
```diff
@@ -0,0 +1,52 @@
...
+        value: iface([], {
+          id: IS_REQ_STRING,
+          hash: IS_REQ_STRING,
+          hashAlg: IS_REQ_STRING,
+        }),
+      }),
+    }),
+
+    UpdateDelegate: iface([], {
+      did: IS_REQ_STRING,
```
JGiter, line 26: Is it possible to specify DID pattern here?

src/modules/didRegistry/didRegistry.validation.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  DIDAttribute,
+  Encoding,
+  PubKeyType,
+} from '@ew-did-registry/did-resolver-interface';
+import { KeyType } from '@ew-did-registry/keys';
+import { createCheckers, iface, enumtype } from 'ts-interface-checker';
+
+const IS_REQ_STRING = 'string';
```
JGiter, line 9: What is the purpose of this alias?

Passerino: as name says if we got 'property': 'string' in interface-checker , it indicates that this field is of type 'string' and it's required. const is not to repeat constant strings across document and avoid typos

src/modules/didRegistry/didRegistry.validation.ts
Passerino: I looked through whole interface and could not find any 'pattern match' or 'regex' or even 'custom validator' .

JGiter (approved):
Thanks. By the way could you remind why we didn't put this in did-reg?

PR #504, opened 2022-03-09
chore: VP POC 

Harasz:
### Contributor checklist


src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -48,3 +49,36 @@ export interface VerifyVerifiableCredentialResults {
...
+
+// Verifiable Presentation Types
+
+export interface SignVerifiablePresentationOptions {
+  verificationMethod?: string;
+  proofPurpose?: string;
+  domain?: Partial<TypedDataDomain>;
+}
+
+export interface SignedVerifiablePresentation {
```
jrhender, line 61: `Signed` is a bit redundant as VerifiablePresentation needs to have a `proof`
```suggestion
export interface VerifiablePresentation {
```

jrhender (commented):
Looks like a good start to the VP, nice work! The important test will be if it can be verified 🤞

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -48,3 +68,40 @@ export interface VerifyVerifiableCredentialResults {
   warnings: string[];
   errors: string[];
 }
+
+// Verifiable Presentation Types
+
+export interface SignVerifiablePresentationOptions {
```
jrhender, line 74: I think it's better to use `ProofOptions`. It's a bit more succinct and aligns with VC terminology and DIDKit terminology https://github.com/spruceid/didkit/blob/305481916434f97183c783106aff6f3f0056486c/lib/web/src/lib.rs#L134 
```suggestion
export interface ProofOptions {
```

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -48,3 +68,40 @@ export interface VerifyVerifiableCredentialResults {
   warnings: string[];
   errors: string[];
 }
+
+// Verifiable Presentation Types
```
jrhender, line 72: I think if it deserves a separator comment like this, I would put them in a separate file.

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -48,3 +68,40 @@ export interface VerifyVerifiableCredentialResults {
...
+
+export interface UnsignedVerifiablePresentation {
+  '@context': string[];
+  id: string;
+  type: 'VerifiablePresentation';
+  holder: string;
+  verifiableCredential: SignedVerifiableCredential<unknown>[];
+}
+
+export interface VerifiablePresentation extends UnsignedVerifiablePresentation {
```
jrhender, line 88: Instead of `UnsignedVerifiablePresentation`, can just say `Presentation`. This is the term VC-API uses for example to refer to a VP that doesn't yet have a proof https://w3c-ccg.github.io/vc-api/holder.html#tag/presentation
```suggestion
export interface VerifiablePresentation extends Presentation {
```

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -48,3 +68,40 @@ export interface VerifyVerifiableCredentialResults {
   warnings: string[];
   errors: string[];
 }
+
+// Verifiable Presentation Types
+
+export interface SignVerifiablePresentationOptions {
+  verificationMethod?: string;
+  proofPurpose?: string;
+  domain?: Partial<TypedDataDomain>;
```
jrhender, line 77: Is `domain` needed? I think we were saying that it isn't necessary because we're not submitting to a smart contract, right?
The signature spec says that
> domain MUST have the value as defined in EIP712. If domain is not provided, a default object is applied with the only property being name with value EthereumEip712Signature2021

https://w3c-ccg.github.io/ethereum-eip712-signature-2021-spec/#suite-definition

```suggestion
```

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
```diff
@@ -1,28 +1,54 @@
...
   VerifyVerifiableCredentialResults,
+  CreateVerifiableCredentialParams,
+  EWFCredentialSubject,
+  VerifiableCredential,
+  UnsignedVerifiablePresentation,
 } from './verifiableCredentials.types';
 
 export abstract class VerifiableCredentialsServiceBase {
-  abstract prepareIssueCredential: (
+  protected abstract prepareIssueCredential: (
```
jrhender, line 22: Maybe my brain isn't working, I don't understand why it would make sense to use `protected` and `abstract` together. `protected` implies an implementation and `abstract` implies no implementation, to me anyways 🤔

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
```diff
@@ -32,67 +58,69 @@ export abstract class VerifiableCredentialsServiceBase {
...
+          : [],
+        id: params.subject,
+      },
+    };
+    return credential;
   }
 
-  public async signVerifiableCredential<T>(
-    credential: VerifiableCredential<T>,
+  // TODO: Check if issuer is allowed to issue given credential
```
jrhender, line 103: I would remove this TBH. IMO, this would imply checking the role definition or something the sort and that would couple this code to a governance framework.
```suggestion
```

jrhender (commented):
Looks solid, added some comments

src/modules/verifiableCredentials/eip712Types.ts
```diff
@@ -0,0 +1,79 @@
+export const verifiableCredentialEIP712Types = {
```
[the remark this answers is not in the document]
jrhender: Sorry, I realized that this might not be a good suggestion if we're moving to Spruce anyways, so wouldn't be worth the effort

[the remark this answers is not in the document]
Harasz: Spruce doesn't generate this, it's also in our hands. I can generate this dynamically, but it is not a easy task as you thinking. That library only generating types for primitive values. We can do that, and i think we will need this, but currently it's better to focus on other thigs.

package.json
```diff
@@ -34,7 +34,7 @@
     "start-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" -p 8544 --chainId 1337 -a 20 -e 10000000 -l=10000000",
     "ganache": "ganache-cli -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" -p 8544 --chainId 1337 -a 20 -e 10000000 -l=10000000",
     "test:watch": "npm run start-rpc -- \"jest --coverage --env=./jest.env.js --watchAll\"",
-    "test": "npm run start-rpc -- \"jest --no-cache --env=./jest.env.js\"",
+    "test": "npm run start-rpc -- \"jest --no-cache -i 'verifiableCredentials.service.e2e.ts' --env=./jest.env.js\"",
```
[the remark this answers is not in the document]
Harasz: That's right :+1:

src/modules/verifiableCredentials/verifiableCredentials.types.ts
Harasz: It's not needed, but i left an option to specify it. You think it's better to remove it?

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
Harasz: Abstract is saying that derived class need to implement such class. Default access modifiers is public, so we could use this method outside our class, but we don;t want to.

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
Harasz: So verifying will need to check it?

src/modules/verifiableCredentials/eip712Types.ts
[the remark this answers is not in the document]
jrhender: Ok, makes sense. Let's keep it as is for now.

src/modules/verifiableCredentials/verifiableCredentials.types.ts
jrhender: Yes, better to remove IMO, because we can always add later but if we want to remove it later, it is a breaking change.

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
jrhender: Ah, makes sense, thanks 👍

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
jrhender: The issuer might also check it, but I don't think it should be done in this location. IMO, as the name is `signVerificableCredential`, function should only do one thing (following the Single Responsibility Principle), which is sign the credential.

src/modules/verifiableCredentials/verifiableCredentials.types.ts
Harasz: Okay, Will be removed

src/modules/verifiableCredentials/verifiableCredentials.types.ts
```diff
@@ -32,19 +51,45 @@ export interface SignedVerifiableCredential<T> extends VerifiableCredential<T> {
...
+      messageSchema: {
+        VerifiablePresentation: TypedDataField[];
+        EIP712Domain: TypedDataField[];
+        Proof: TypedDataField[];
+        VerifiableCredential?: TypedDataField[];
+      };
+      primaryType: string;
+    };
+  };
+}
```
jrhender, line 95: Instead of our own types, could we import from PEX? I think we'll probably be using PEX in ICL, so I don't think it will add additional weight to the library https://github.com/Sphereon-Opensource/pex/blob/c23a250c8581c76fbefac59fca90b418bbe19c2f/lib/types/SSI.types.ts#L104 
I think PEX doesn't have some of the special eip712Domain details, but perhaps that isn't a problem or can be added via another interface?

jrhender (changes requested):
Thanks for all of your work on this. I left some additional comments.

Another thing I noticed is that the module directory is `verifiableCredentials`. By our agreed upon naming convention, shouldn't it be `verifiable-credentials`?

src/modules/verifiableCredentials/verifiableCredentials.types.ts
Harasz: Okay, if there is an intention to use PEX in ICL in the future i think it's not a problem.

Harasz:
@jrhender I agree, but as I can see ICL has currently a mess in name convention. We should create a task to rename all modules and files to match our new convention. I will rename VC/VP module

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
```diff
@@ -131,6 +159,125 @@ export abstract class VerifiableCredentialsServiceBase {
       signature
     );
 
-    return JSON.parse(signedCredential) as SignedVerifiableCredential<T>;
+    return JSON.parse(
+      signedCredential
+    ) as VerifiableCredential<EWFCredentialSubject>;
+  }
+
+  public createPresentation(
```
jrhender, line 167: Speaking of PEX, instead of this function, could we use `pex.presentationFrom` instead? https://github.com/Sphereon-Opensource/pex#holder-presentation-creation-non-verifiable

jrhender (changes requested):

src/modules/verifiableCredentials/verifiableCredentialsBase.service.ts
Harasz: Definitely! I will try to use as many features from PEX as I can

Harasz: Problem is that this methods are doing different things. Method that I implemented is creating from a given VCs, but PEX require VCs and presentation definition. Do we want to replace my method with that from PEX or maybe we can create another method

jrhender: Ah, I see, the problem is that PEX method requires a `presentationDefinition`. Hmm. I think we should definitely use the pex method as . However, I think it would be code to encapsulate it so that we can change the implementation or support other presentation types. At the moment, my suggestion is an API like this:
```
createPresentation(verifiableCredentials: VerifiableCredential<EWFCredentialSubject>[], options: createPresentationOptions)
```
Where `createPresentationOptions` looks like this:
```
{
  type: PresentationType, // e.g. "PresentationExchangeSubmission"
  presentationDefinition?: PresentationDefinition
}
```
Or maybe it's simpler just to do
```
createPEXPresentation(verifiableCredentials: VerifiableCredential<EWFCredentialSubject>[], presentationDefinition: PresentationDefinition)
```
What do you think?

src/modules/verifiable-credentials/verifiable-credentials-node.service.ts
```diff
@@ -0,0 +1,61 @@
...
+} from 'didkit-wasm-node';
+import {
+  prepareIssuePresentation,
+  completeIssuePresentation,
+} from '@spruceid/didkit-wasm-node';
+import { SignerService } from '../signer';
+import { VerifiableCredentialsServiceBase } from './verifiable-credentials-base.service';
+
+export class VerifiableCredentialsServiceNode extends VerifiableCredentialsServiceBase {
+  protected prepareIssueCredential: (
```
JGiter, line 15: why duplicate declarations in sub-classes?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (
+      vp.proof.eip712Domain.primaryType === 'VerifiablePresentation' &&
+      'verifiableCredential' in vp &&
+      vp.verifiableCredential
+    ) {
+      const vcsVerification = await Promise.all([
+        ...vp.verifiableCredential.map((vc) => this.verify(vc)),
+      ]);
+
+      return true && vcsVerification.every(Boolean);
```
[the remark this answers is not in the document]
Harasz: It should be like that :smile: Will fix

src/modules/verifiable-credentials/verifiable-credentials-node.service.ts
Harasz: To have ability to use here the function from DIDKit `this.prepareIssueCredential = prepareIssueCredential`

JGiter: Isn't possible to override them without re-declaration like
```
prepareIssueCredential(...) {
   return prepareIssueCredential(...)
}
```

Harasz: It is possible. But why this is better then simple overriding?

JGiter: Overriding is happen in any case either as method or property overriding, but method overriding allows to avoid re-declaration

Harasz: I don't think so. It's extending an abstract method, so it have to be declared in derived class

JGiter: ok

PR #507, opened 2022-03-16
task/ICL-258 get published offchain claims

JGiter:
### Description

https://energyweb.atlassian.net/browse/ICL-258

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -7,7 +7,10 @@ import { Methods } from '@ew-did-registry/did';
 import { Algorithms } from '@ew-did-registry/jwt';
 import { addressOf } from '@ew-did-registry/did-ethr-resolver';
 import { hashes, IPublicClaim } from '@ew-did-registry/claims';
-import { DIDAttribute } from '@ew-did-registry/did-resolver-interface';
+import {
+  DIDAttribute,
+  IServiceEndpoint,
+} from '@ew-did-registry/did-resolver-interface';
 import { ClaimManager__factory } from '../../../ethers/factories/ClaimManager__factory';
 import { ERROR_MESSAGES } from '../../errors';
 import { emptyAddress } from '../../utils/constants';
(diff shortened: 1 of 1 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -590,14 +594,24 @@ export class ClaimsService {
...
-  }: { did?: string } | undefined = {}) {
+  }: { did?: string } | undefined = {}): Promise<Claim[]> {
     const { service } = (await this._didRegistry.getDidDocument({ did })) || {};
-    return service;
+    const issuedClaims = await this.getClaimsBySubject({
+      did,
+      isAccepted: true,
+    });
+    return issuedClaims.filter((c) =>
+      service.some(
```
Harasz, line 609: This logic is a little bit fuzzy for me. Could you explain what you are trying to achieve here?

JGiter: Sure. At the moment getUserClaims actually returns service endpoints. So this change is aimed to return published issued claims, which is ensured by checking that for issued offchain claim exists corresponding service. I would also change method name to getOffchainClaims({isApproved, isPublished}) or even better extend getClaimsBy.. with registrationTypes parameter if this wasn't breaking change affecting a lot of code

Harasz: Firstly, `service` variable can be undefined, so `service.some()` can throw an error.
And what about something like this
 ```suggestion
      const { service } = (await this._didRegistry.getDidDocument({ did })) || {};

if (!service) return []

const issuedClaims = await this.getClaimsBySubject({
  did,
  isAccepted: true,
});

const issuedClaimsTypes = issuedClaims.map(({ claimType }) => claimType);

return issuedClaims.filter(
  ({ registrationTypes, claimType }) =>
    registrationTypes.includes(RegistrationTypes.OffChain) &&
    issuedClaimsTypes.includes(claimType)
);
```

I think it's the code is more readable

JGiter: I am agree with null checking of `service`, but filtering of issued claims is unclear to me. It doesn't seems that filtering excludes not-published claims (not added to document). Also second condition seems like holds true for all claims

Harasz: I agree, `issuedClaimsTypes` should map a claim type from `service` array, then everything should be fine, what you think?

JGiter: Seems fine to me

Harasz (approved):

Harasz (approved):

PR #511, opened 2022-03-23
feat(core): feat/ICL-114 custom logger

Passerino:
all console logs turned into logger implementing ILogger
functionalities:
- default logger - ConsoleLogger
- ILogger : debug, info, warn, error
- ability to get logger everywhere with getLogger():ILogger
- ability to set custom logger implmenting ILogger with setLogger(ILogger)
- ability to set min logLevel

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -1,0 +1,46 @@
+export enum LogLevel {
+  debug = 1,
+  info = 2,
+  warn = 3,
+  error = 4,
+}
+
+/*
+ * Abstract overridable class for logger
+ */
+export abstract class ILogger {
+  constructor(protected readonly _logLevel: LogLevel = LogLevel.debug) {}
+
+  debug = (message: unknown) => this.log(message, LogLevel.debug);
+  info = (message: unknown) => this.log(message, LogLevel.info);
+  warn = (message: unknown) => this.log(message, LogLevel.warn);
+  error = (message: unknown) => this.log(message, LogLevel.error);
+
+  log = (message: unknown, level: LogLevel) =>
+    this._logLevel <= level && this._log(message, level);
+
+  protected abstract _log(message: unknown, level: LogLevel): void;
+}
+
+/*
+ * Default Implementation of ILogger streaming logs to console
+ */
+export class ConsoleLogger extends ILogger {
+  _log = (message: unknown, level: LogLevel) => {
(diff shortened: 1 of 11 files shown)


src/utils/logger.ts
```diff
@@ -0,0 +1,39 @@
...
+  debug = 1,
+  log = 2,
+  warn = 3,
+  error = 4,
+}
+
+/*
+ * Default Implementation of ILogger streaming logs to console
+ */
+export class ConsoleLogger extends ILogger {
```
JGiter, line 23: A small optimization: we can filter logs in base class and in subclasses implement only specific logger feature. For example `ILogger.warn` could be
`if (this._logLevel <=LogLevel.warn) {
   this.log(message)
`
So the only abstract method is `log`. Also method currently name `log` can be renamed `info`

JGiter (changes requested):

src/utils/logger.ts
Passerino: this would require to duplicate all functions but I'm up for that. We would have 
- 4 non abstract functions : `info(), warn(), debug(), error()` containing log level filter and 
- 4 abstract private functions `logInfo(), logWarn(), logDebug(), logError()` containing only logging functionality.   

I this how you imagin, too ?

JGiter: Not actually. I mean there is only one abstract method - the one which is actually does logging. And only this method is overridden in sub-classes. Like this:
```
abstract class ILogger {
   abstract _log(message); // only this is overridden

   info(message) {
      if (this._logLevel <= LogLevel.Info) {
         this.**_log**(message);
      }
   }

   warn(message) {
      if (this._logLevel <= LogLevel.Warn) {
         this.**_log**(message);
      }
   }
  ...
}

class ConsolLoger extends ILogger {
   _log(message) {
       console.log(message);
   }
}

```

JGiter (changes requested):

src/utils/logger.ts
Passerino: but how would we know if this is error or warn or anything else if the only thing we always log is  log(message). They usually differ with format , eg `{LogLevel} - {date} - {message}` , sometimes `color` how would you achieve it with your solution?

JGiter (approved):
Thanks! This looks good

PR #514, opened 2022-03-30
doc(core): custom logger docs

Passerino:
### Description

Doc about logging in IAM Client Lib and how to overide it with custom logger


jrhender (approved):
Thanks @Passerino

PR #519, opened 2022-04-07
task/ICL-276 initiate verifiable credentials exchange

JGiter:
### Contributor checklist


src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,30 @@
...
+  PresentationDefinition,
+  VerifiableCredential,
+  CredentialSubject,
+} from '@ew-did-registry/credentials-interface';
+import { AxiosInstance } from 'axios';
+import { CacheClient } from '../cacheClient';
+
+export default class VCStorageClient {
+  private _http: AxiosInstance;
+  private _baseUrl = 'vc';
```
Harasz, line 11: We will be using `vp` and `webnode` prefix, so I think we can remove  this parameter

src/modules/verifiable-credentials/types/role-credential.types.ts
```diff
@@ -1,4 +1,8 @@
-import { CredentialSubject } from '@ew-did-registry/credentials-interface';
+import {
+  CredentialSubject,
+  VerifiableCredential,
+} from '@ew-did-registry/credentials-interface';
+import { has} from 'lodash';
```
Harasz, line 5:
```suggestion
import { has } from 'lodash';
```

Harasz (changes requested):

src/modules/claims/claims.service.ts
```diff
@@ -517,6 +517,7 @@ export class ClaimsService {
     registrationTypes?: RegistrationTypes[];
     claim: { token?: string; claimType?: string };
   }) {
+    console.log('claim to publish:', claim);
```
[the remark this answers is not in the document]
JGiter: will remove

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(cacheClient: CacheClient) {
+    this._http = cacheClient.http;
+  }
+
+  async getCredentialsByDefinition(
+    definition: PresentationDefinition
+  ): Promise<VerifiableCredential<CredentialSubject>[]> {
+    const { data } = await this._http.get<
+      VerifiableCredential<CredentialSubject>[]
+    >(`/match`, { baseURL: 'vp', params: definition });
```
Harasz, line 21: https://github.com/energywebfoundation/ssi-hub/blob/ba3ea1edf48ab1f433c92ad4d13131a7845cad2e/src/modules/presentation-exchange/presentation-exchange.controller.ts#L23
It should be a post method with a body

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,48 @@ export abstract class VerifiableCredentialsServiceBase {
...
+        presentationDefinitions.map((def) =>
+          this._storage.getCredentialsByDefinition(def)
+        )
+      )
+    ).reduce((prev, curr) => prev.concat(curr), []);
+    for (const cred of credentials) {
+      if (isRoleCredential(cred)) {
+        vc.push(cred);
+      }
+    }
```
Harasz, line 135:
```suggestion
    )
    .reduce((prev, curr) => prev.concat(curr), [])
    .filter(isRoleCredential);
```
And `credentials` array is the same as `vc` should be. We get rid of one variable and 3 lines of code. What you think?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,48 @@ export abstract class VerifiableCredentialsServiceBase {
...
+    const vp = await this.createVerifiablePresentation(vc);
+    return { vp, vpRequest };
+  }
 
   // * Should be overridden by the implementation
   static async create(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    signerService: SignerService
+    signerService: SignerService,
+    storage: VCStorageClient
```
Harasz, line 145:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    storage: VCStorageClient
```

Harasz (changes requested):
You have changes from this PR here. https://github.com/energywebfoundation/iam-client-lib/pull/522

Lets rebase it with develop first

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default class VCStorageClient {
+  private _http: AxiosInstance;
+
+  constructor(cacheClient: CacheClient) {
+    this._http = cacheClient.http;
+  }
+
+  async getCredentialsByDefinition(
+    definition: PresentationDefinition
```
jrhender, line 17: I think this would be a bit more accurate
```suggestion
  async getCredentialsByPresentationDefinition(
    presentationDefinition: PresentationDefinition
```

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,43 @@ export abstract class VerifiableCredentialsServiceBase {
...
   ): Promise<string>;
 
-  constructor(protected readonly _signerService: SignerService) {}
+  constructor(
+    protected readonly _signerService: SignerService,
+    private readonly _storage: VCStorageClient
+  ) {}
+
+  /**
+   * @description requests verifiable credentials of type `type`. Only EWF credentials currently supported
```
jrhender, line 110: The type that we're currently supporting is `vc-api-exchange`. This is independent from EWF credentials.
```suggestion
   * @description The type of the exchange. Only vc-api exchanges currently supported. 
```

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,43 @@ export abstract class VerifiableCredentialsServiceBase {
...
+   * @returns credentials query with matching verifiable presentations
+   */
+  async initiateExchange({ type, url }: ExchangeInvitation) {
+    if (type !== EWF_EXCHANGE_INVITATION) {
+      throw new Error('Only EWF exchange is supported');
+    }
```
jrhender, line 116:
```suggestion
    if (type !== VC_API_EXCHANGE) {
      throw new Error('Only VC-API exchange is supported');
    }
```

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,43 @@ export abstract class VerifiableCredentialsServiceBase {
...
+
+    const vc: VerifiableCredential<RoleCredentialSubject>[] = (
+      await Promise.all(
+        presentationDefinitions.map((def) =>
+          this._storage.getCredentialsByDefinition(def)
+        )
+      )
+    )
+      .reduce((prev, curr) => prev.concat(curr), [])
+      .filter(isRoleCredential);
```
jrhender, line 131: I don't think restricting this to `RoleCredentials` is necessary, unless I'm missing something
```suggestion
    const vc: VerifiableCredential<CredentialSubject>[] = (
      await Promise.all(
        presentationDefinitions.map((def) =>
          this._storage.getCredentialsByDefinition(def)
        )
      )
    )
      .reduce((prev, curr) => prev.concat(curr), []);
```

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,43 @@ export abstract class VerifiableCredentialsServiceBase {
...
+  /**
+   * @description requests verifiable credentials of type `type`. Only EWF credentials currently supported
+   * @returns credentials query with matching verifiable presentations
+   */
+  async initiateExchange({ type, url }: ExchangeInvitation) {
+    if (type !== EWF_EXCHANGE_INVITATION) {
+      throw new Error('Only EWF exchange is supported');
+    }
+
+    const { data: vpRequest } = await axios.post<VpRequest>(url);
```
jrhender, line 118: Should this type be `VpRequest` with capital `V`?
```suggestion
    const { data: VpRequest } = await axios.post<VpRequest>(url);
```

jrhender (approved):
Looks like a great start, thanks @JGiter . I added some comments but if it looks good from @Harasz 's end, we could merge IMO

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: It's not a type, it's an alias :face_in_clouds:

jrhender: Ah right! I didn't recognize it as that, thanks!

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,25 @@
...
+  ): Promise<VerifiableCredential<CredentialSubject>[]> {
+    const { data } = await this._http.post<
+      VerifiableCredential<CredentialSubject>[]
+    >(`/match`, { baseURL: 'vp', params: definition });
```
Harasz, line 21:
```suggestion
    >(`/vp/match`, definition);
```
We can't override `baseUrl` here, bc it's actually overriding SSI-Hub URL. Also `definition` should be in body not in params.

Harasz (changes requested):

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
JGiter: This is because we create presentation of role credentials https://github.com/energywebfoundation/iam-client-lib/blob/b96d23a4c0ed021a7c229ea39ecfd4b03cca0849/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts#L197

e2e/verifiable-credentials.service.e2e.ts
```diff
@@ -220,4 +285,29 @@ describe('Verifiable credentials tests', () => {
...
+    test('initiateExchange() should return presentation matching presentation request', async () => {
+      const vc = await createExampleSignedCredential([]);
+      (axios as jest.Mocked<typeof axios>).post.mockImplementationOnce(
+        (url) => {
+          return Promise.resolve({
+            data: url === exchangeUrl ? vpRequest : '',
+          });
+        }
+      );
+      mockGetCredentialsByDefinition.mockImplementationOnce((vpDefinition) =>
```
Harasz, line 301: I think this is not good idea to mock also storage client methods. Mocking `axios` should be enough

JGiter: I would prefer to mock only direct dependencies of storage-client. Otherwise we will need to change tests if implementation of storage-client changed

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
JGiter: I think we can delete this method at all. Why do we need it?

jrhender:
@JGiter Could you resolve the merge conflicts when you have time?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -93,12 +101,43 @@ export abstract class VerifiableCredentialsServiceBase {
...
+      await Promise.all(
+        presentationDefinitions.map((def) =>
+          this._storage.getCredentialsByPresentationDefinition(def)
+        )
+      )
+    )
+      .reduce((prev, curr) => prev.concat(curr), [])
+      .filter(isRoleCredential);
+    const vp = await this.createVerifiablePresentation(vc);
+    return { vp, vpRequest };
```
jrhender, line 133: In my opinion, this function should not return the actual VP but should instead an object which contains possible credentials for selection by the user in the wallet.
I'll make a small PR that can demonstrate this.

jrhender, line 133: Instead of returning the VCs here, we can return the `SelectResult` and `PresentationDefinition` for Switchboard to display to user for selection and confirmation
```suggestion
    return await Promise.all(
      presentationDefinitions.map(async (presentationDefinition) => {
        const selectResults = await this._storage.getCredentialsByPresentationDefinition(presentationDefinition);
        return {
          presentationDefinition,
          selectResults
        }
      })
    )
```

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  async getCredentialsByPresentationDefinition(
+    definition: PresentationDefinition
+  ): Promise<VerifiableCredential<CredentialSubject>[]> {
+    const { data } = await this._http.post<
+      VerifiableCredential<CredentialSubject>[]
+    >(`/vp/match`, definition);
+
+    return data;
```
jrhender, line 23: I think this section is incorrect. I think that `/vp/match` is returning the a `MatchResultsDto` which has more than just `VerifiableCredentials`. 
https://identitycache-dev.energyweb.org/api/#/Verifiable%20Presentation/PresentationExchangeController_getMatchedCredentialsForVpDefinition
```suggestion
  async getCredentialsByPresentationDefinition(
    definition: PresentationDefinition
  ): Promise<SelectResults> {
    const { data } = await this._http.post<
      SelectResults
    >(`/vp/match`, definition);

    return data;
  }
```

jrhender (changes requested):
`initiateExchange` doesn't need to create the VP. Instead, the `SelectResults` and `PresentationDefinition` should be passed back to the wallet for the user to review and confirm

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,23 @@
+import { PresentationDefinition } from '@ew-did-registry/credentials-interface';
```
Harasz, line 1: Did you remove my changes that I've made here #523?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: We can create a static abstract method, and this is making sure that derived class has implemented this method

Harasz (approved):

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: In ES6 it is possible

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -0,0 +1,113 @@
...
+  ): Promise<SelectResults> {
+    const { data } = await this._http.post<SelectResults>(
+      `/vp/match`,
+      definition
+    );
+
+    return data;
+  }
+
+  async store(
```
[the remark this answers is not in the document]
jrhender: Removed the documentation suggestion as these aren't actually exported from the module, so won't be seen by consumers of iam-client-lib

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: I've already done it. It will throw an error on runtime if derived class doesn't have this method

src/init.ts
```diff
@@ -115,13 +118,13 @@ export async function init(signerService: SignerService) {
       assetsService,
       connectToDidRegistry,
       stakingPoolService,
+      verifiableCredentialsService,
```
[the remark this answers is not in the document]
jrhender: @JGiter Were you able to update the `quick_start` doc with this new service? Or were you thinking we should do it in a separate story/task? (Or did I just miss it in the changes 😅 )

jrhender (approved):
Looks good, maybe just updating the `quick_start.md` with how to get this new service

src/init.ts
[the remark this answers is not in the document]
JGiter: Updated now. I skipped updating before because I thought you removed documentation suggestion about verifiable credential service. Probably it was about storage client

PR #522, opened 2022-04-12
fix(claim): fix public claim publishing

Harasz:
### Description

Fix role publishing for on chain claim

### Contributor checklist

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -521,16 +521,21 @@ export class ClaimsService {
     this.validatePublishPublicClaimRequest(registrationTypes, claim);
     let url: string | undefined = undefined;
     if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+      if (!claim.claimType) {
+        throw new Error(ERROR_MESSAGES.CLAIM_TYPE_MISSING);
+      }
+
       const claims = await this.getClaimsBySubject({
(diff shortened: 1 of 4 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -521,16 +521,22 @@ export class ClaimsService {
...
     if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+      if (!claim.claimType) {
+        throw new Error(ERROR_MESSAGES.CLAIM_TYPE_MISSING);
+      }
+
       const claims = await this.getClaimsBySubject({
         did: this._signerService.did,
-        namespace: claim.claimType,
+        // get namespace from claimType
+        namespace: claim.claimType.split('.').slice(2).join('.'),
```
dawidgil, line 531: isn't it better to create method or function `getNamespaceFromClaimType(claimType: string): string` ? This way there is no need to create a comment.

dawidgil (approved):

src/modules/claims/claims.service.ts
jrhender, line 531: This might be a bit clearer to readers as it makes it clear we want everything before `.roles.`. What do you think?

```suggestion
        namespace: claim.claimType.split('.roles.')[1],
```

JGiter (approved):

PR #523, opened 2022-04-12
feat(vc): store vc in decentralized-web-node

Harasz:
### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/storage-client.ts
+++ b/src/modules/verifiable-credentials/storage-client.ts
@@ -4,13 +4,28 @@ import {
   CredentialSubject,
 } from '@ew-did-registry/credentials-interface';
 import { AxiosInstance } from 'axios';
+import { CID } from 'multiformats/cid';
+import * as json from 'multiformats/codecs/json';
+import { sha256 } from 'multiformats/hashes/sha2';
+import { v4 } from 'uuid';
 import { CacheClient } from '../cacheClient';
+import { SignerService } from '../signer';
+import {
+  WebNodeRequestObject,
+  WebNodeResponseObject,
+  StoreVcResult,
+  WebNodeWriteMessage,
+  WebNodeQueryMessage,
+  WebNodeReply,
+} from './types';
 
 export default class VCStorageClient {
-  private _http: AxiosInstance;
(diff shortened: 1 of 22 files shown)


src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -22,4 +32,80 @@ export default class VCStorageClient {
...
+    if (response?.status?.code && response.status.code >= 400) {
+      throw new Error(response.status.message || 'Unknown error');
+    }
+
+    if (!response?.replies) {
+      throw new Error('No replies in response');
+    }
+
+    return response?.replies.map((replay) => {
+      if (replay?.status?.code && replay.status.code >= 400) {
```
jrhender, line 91: Minor typo, should be `reply`
```suggestion
    return response?.replies.map((reply) => {
      if (reply?.status?.code && reply.status.code >= 400) {
```

src/modules/verifiable-credentials/storage-client.ts
```diff
@@ -22,4 +32,80 @@ export default class VCStorageClient {
...
+      requestId: v4(),
+      target: 'vp',
+      messages,
+    };
+
+    const { data: response } = await this._http.post<
+      WebNodeResponseObject<[WebNodeWriteMessage]>
+    >(`/`, request, {
+      baseURL: 'webnode',
+    });
```
jrhender, line 80: How does this work with the `ssi-hub` authentication. I think it might work if using browser cookies, but the I think the axios instance in `CacheClient` potentially has an authentication token which it is sending as a Bearer Token. So maybe this call should be in the `CacheClient`?

jrhender: Ah sorry, I missed that `CacheClient` is coming in via the constructor. I think this is fine then

Harasz: We are using `axios` instance from `CacheClient`, so `CacheClient` is taking care of authentication. It's including cookies and headers to each request including to this from `webnode`

JGiter (approved):

JGiter (approved):

PR #524, opened 2022-04-14
task/ICL-277 continue credential exchange

JGiter:
### Description

https://energyweb.atlassian.net/browse/ICL-277


jrhender:
@JGiter Could you rebase with develop when you have time please? That will make it easier to review 🙇

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -108,7 +111,12 @@ export abstract class VerifiableCredentialsServiceBase {
    * @description The type of the exchange. Only vc-api exchanges currently supported.
    * @returns credentials query with matching verifiable presentations
    */
-  async initiateExchange({ type, url }: ExchangeInvitation) {
+  async initiateExchange({ type, url }: ExchangeInvitation): Promise<
+    {
+      selectResults: SelectResults;
+      presentationDefinition: PresentationDefinition;
+    }[]
```
jrhender, line 118: Nice. Personally, I'm a fan of having the return type on the method signature

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -132,6 +140,44 @@ export abstract class VerifiableCredentialsServiceBase {
...
+      credentials,
+      {
+        challenge,
+      }
+    );
+
+    const {
+      data: { errors, vp: requestedPresentation, vpRequest: nextVpRequest },
+    } = await axios.put<ExchangeResponse>(service[0].serviceEndpoint, {
+      requestedPresentation: requiredPresentation,
```
jrhender, line 169: Would it make sense to align the names here?
```suggestion
      requiredPresentation: requiredPresentation,
```

jrhender (commented):
Looks like a good start!

jrhender:
@JGiter Let me know when you want review again (please re-request)

src/modules/verifiable-credentials/types/verifiable-credentials.types.ts
```diff
@@ -18,6 +18,7 @@ export interface RoleCredentialSubjectParams {
 export interface ProofOptions {
   verificationMethod?: string;
   proofPurpose?: string;
+  challenge: string;
```
Harasz, line 21:
```suggestion
  challenge?: string;
```
I don't think we need `challenge` for all cases

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: I am also not a fan of inline types

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -132,6 +140,44 @@ export abstract class VerifiableCredentialsServiceBase {
...
+  public async continueExchange({
+    vpRequest: {
+      challenge,
+      interact: { service },
+    },
+    credentials,
+  }: {
+    vpRequest: VpRequest;
+    credentials: VerifiableCredential<RoleCredentialSubject>[];
+  }) {
```
Harasz, line 158: Move inline type to interface in separate file

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
JGiter: Actually, it looks like the right name is just `presentation` as it used in https://github.com/energywebfoundation/ssi/blob/master/apps/vc-api/docs/tutorials/rebeam-tutorial.md#customer-create-the-required-presentation. Is that right?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -108,18 +113,21 @@ export abstract class VerifiableCredentialsServiceBase {
...
+    type,
+    url,
+  }: ExchangeInvitation): Promise<ContinueExchangeSelections> {
     if (type !== VC_API_EXCHANGE) {
       throw new Error('Only VC-API exchange is supported');
     }
 
     const { data: vpRequest } = await axios.post<VpRequest>(url);
-    const presentationDefinitions = vpRequest.query.find(
+    const credentialQuery = vpRequest.query.find(
```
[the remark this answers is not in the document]
JGiter: I am agree we should verify this. But we can go even further and validate whole vpRequest with class-validator. Do you think it is worth it or on this stage it is enough null checks of query and credentialQuery?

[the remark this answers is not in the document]
whitneypurdum: Oh I think up to you. I think even if we use class-validator, it would still throw an error here just from doing a .find on undefined or null. Sorry if I'm misunderstanding your question.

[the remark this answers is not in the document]
JGiter: Yes, it will still throw an error. The benefit is that several null-checks will be replaced with one validation

PR #541, opened 2022-04-28
docs(did-registry): improve DID registry service documentation

Harasz:
### Description

- Improve did registry service methods documentation
- Add examples to did registry service methods
- Move inline type to interfaces

### Contributor checklist


src/modules/claims/claims.service.ts
```diff
@@ -575,25 +575,26 @@ export class ClaimsService {
...
-      return id;
+      return service.id;
     }
 
     if (
       claimData.claimType &&
-      id &&
-      claimData.claimTypeVersion === claimTypeVersion
+      service.id &&
+      claimData.claimTypeVersion === service.claimTypeVersion
```
JGiter, line 595: May be we remove this now? From the way the service was found implies that this conditions are satisfied

JGiter (approved):

src/modules/claims/claims.service.ts
Harasz: I don't want to change the logic here, maybe just little refactors (that's all). So create a ticket in jira pls

PR #547, opened 2022-05-10
task/ICL-210 verify EWF verifiable credential claim

JGiter:
### Description

https://energyweb.atlassian.net/browse/ICL-210

merge after https://github.com/energywebfoundation/ew-credentials/pull/17


e2e/claims.service.e2e.ts
```diff
@@ -43,14 +44,14 @@ const roleName2 = 'myrole2';
 const roleName3 = 'myrole3';
 const roleName4 = 'myrole4';
 const namespace = root;
-const version = 1;
+const version = '1';
```
Harasz, line 47: you need `int` type only in one place, why then use `parseInt()` in a couple of places instead of  `version.toString()` in one place?

src/modules/claims/claims.service.ts
```diff
@@ -90,6 +96,31 @@ export class ClaimsService {
...
+        return (<IRoleDefinition>def).issuer;
+      },
+    };
+    this._issuerVerifier = new IssuerVerification(
+      this._signerService.provider,
+      this._didRegistry.registrySettings,
+      credentialResolver,
+      issuerResolver
+    );
   }
```
Harasz, line 124: Maybe move that to function called `setIssuerVerifier` or something like that

src/modules/claims/claims.service.ts
```diff
@@ -578,6 +608,16 @@ export class ClaimsService {
     return message.issuedToken;
   }
 
+  /**
+   * Verifies Energy Web role verifiable credential and issuer authorization
+   *
+   * @param claim Verifiable credential of Energy Web role
+   */
+  async verifyClaimVc(claim: VerifiableCredential<RoleCredentialSubject>) {
```
Harasz, line 616:
```suggestion
  async verifyClaimVc(claim: VerifiableCredential<RoleCredentialSubject>): Promise<void> {
```

src/modules/claims/claims.service.ts
```diff
@@ -578,6 +608,16 @@ export class ClaimsService {
     return message.issuedToken;
   }
 
+  /**
+   * Verifies Energy Web role verifiable credential and issuer authorization
+   *
+   * @param claim Verifiable credential of Energy Web role
```
Harasz, line 614:
```suggestion
   * @param {VerifiableCredential<RoleCredentialSubject>} claim Verifiable credential of Energy Web role
```

Harasz (approved):

jrhender:
@JGiter do you know why the tests are failing here? I guess you're still working on them?

JGiter:
Tests are failing now because https://github.com/energywebfoundation/ew-credentials/pull/17 isn't merged yet

src/modules/claims/claims.service.ts
```diff
@@ -441,6 +452,19 @@ export class ClaimsService {
     await this._cacheClient.issueClaim(this._signerService.did, message);
   }
 
+  async verifyVc(vc: VerifiableCredential<RoleCredentialSubject>) {
```
Harasz, line 455: Missing documentation

src/modules/claims/claims.service.ts
```diff
@@ -1114,15 +1138,16 @@ export class ClaimsService {
   }
 
   /**
-   * Verify if the user is able to issue the given role. Throws an error when the user is not able to issue the given role.
+   * Verify if the user is issuer of the role verifiable credential
    *
    * @param {String} role Registration types of the claim
    */
```
Harasz, line 1144: wrong documentation

src/modules/claims/claims.service.ts
```diff
@@ -1312,4 +1337,22 @@ export class ClaimsService {
       await this._signerService.signMessage(arrayify(proofHash))
     );
   }
+
+  private _setClaimIssuerVerifier() {
```
Harasz, line 1341: Documentation

Harasz (changes requested):

package.json
```diff
@@ -63,22 +63,23 @@
...
-    "@ew-did-registry/jwt": "^0.6.3-alpha.596.0",
-    "@ew-did-registry/keys": "^0.6.3-alpha.596.0",
-    "@ew-did-registry/proxyidentity": "^0.6.3-alpha.596.0",
-    "@ew-did-registry/revocation": "^0.6.3-alpha.596.0",
+    "@ew-did-registry/claims": "^0.6.3-alpha.626.0",
+    "@ew-did-registry/credentials-interface": "^0.6.3-alpha.626.0",
+    "@ew-did-registry/did": "^0.6.3-alpha.626.0",
+    "@ew-did-registry/did-document": "^0.6.3-alpha.626.0",
+    "@ew-did-registry/did-ethr-resolver": "^0.6.3-alpha.626.0",
+    "@ew-did-registry/did-ipfs-store": "^0.6.3-alpha.626.0",
```
whitneypurdum, line 77: @JGiter @nichonien I updated all ew-did-registry and ew-credentials to latest versions

src/modules/claims/claims.service.ts
```diff
@@ -1320,4 +1346,24 @@ export class ClaimsService {
...
+    const credentialResolver: CredentialResolver = new IpfsCredentialResolver(
+      this._signerService.provider,
+      this._didRegistry.registrySettings,
+      this._didRegistry.ipfsStore
+    );
+    const domainReader = new DomainReader({ensRegistryAddress: chainConfigs()[this._signerService.chainId].ensResolverV2Address, provider: this._signerService.provider})
```
whitneypurdum, line 1356: Previous code is commented out for reference to old changes

nichonien, line 1356:
```suggestion
    const domainReader = new DomainReader({ensRegistryAddress: chainConfigs()[this._signerService.chainId].ensRegistryAddress, provider: this._signerService.provider})
```

nichonien: @whitneypurdum @JGiter Am I right with the above suggested change? Also, we need to `addKnownResolver` for the domainReader to be able to resolve namespaces.

JGiter: Yes, both your comments are correct. Actually I think that instead of creating domain reader here it should be exported from domains service or created by issuer verifier

nichonien: Yeah, exporting from `DomainService` sounds good to me since we already have it at `ClaimService` creation https://github.com/energywebfoundation/iam-client-lib/blob/4f3688e5329ec1ccb222a90049583c6633922fb3/src/modules/claims/claims.service.ts#L87

src/modules/claims/claims.service.ts
whitneypurdum: @JGiter I added documentation for this, let me know if it is not what you would have put

src/modules/claims/claims.service.ts
whitneypurdum: @JGiter same for this

src/modules/claims/claims.service.ts
JGiter: We can also stress out that this method verifies offchain credential
``` Verifies offchain role credential according with role definition```

jrhender (commented):
As discussed, I think we should 
1. keep the changes to `verifyVcIssuer` in `claims.service` that we have now
2. move `verifyVc` to a separate PR (and to verifiable-credentials.service)

src/modules/claims/claims.service.ts
nichonien: @whitneypurdum If you could make the above suggested changes, it would be great.

whitneypurdum: Sure thing @nichonien

PR #549, opened 2022-05-11
docs(signer): improve signer service documentation

Harasz:
### Description

- Improve signer service methods documentation
- Add examples to signer service methods
- Move inline type to interfaces

### Contributor checklist

### DIFF
--- a/src/modules/signer/signer.service.ts
+++ b/src/modules/signer/signer.service.ts
@@ -34,6 +34,15 @@ const {
   verifyMessage,
 } = utils;
 export type ServiceInitializer = () => Promise<void>;
+
+/**
+ * Service responsible for signing messages and sending transactions to the blockchain
+ *
+ * ```typescript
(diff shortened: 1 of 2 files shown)


src/modules/signer/signer.service.ts
```diff
@@ -202,13 +327,18 @@ export class SignerService {
   }
 
   /**
-   * @description Tries to create `eth_sign` conformant signature (https://eth.wiki/json-rpc/API#eth_sign)
+   * Tries to create `eth_sign` conformant signature (https://eth.wiki/json-rpc/API#eth_sign).
    * Whether or not to hash the message prior to signature is determined by signature performed during login.
```
JGiter, line 331:
```suggestion
   * Whether or not to hash the message prior to signature is depends on whether is signer EIP-191 compliant.
```

JGiter (approved):

PR #551, opened 2022-05-12
feat(vc): add verifiable credential expiration date

Harasz:
### Description

Add expiration feature to VC and VP implementation 

### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
+++ b/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
@@ -19,7 +19,6 @@ import {
   RoleCredentialSubject,
   CreatePresentationParams,
   verifiableCredentialEIP712Types,
-  verifiablePresentationEIP712Types,
   verifiablePresentationWithCredentialEIP712Types,
   InitiateExchangeResults,
 } from './types';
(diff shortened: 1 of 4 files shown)


src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -300,9 +308,33 @@ export abstract class VerifiableCredentialsServiceBase {
...
     // TODO: generate types from the presentation
-    const types = verifiableCredential
-      ? verifiablePresentationWithCredentialEIP712Types
-      : verifiablePresentationEIP712Types;
+    const types: typeof verifiablePresentationWithCredentialEIP712Types =
+      JSON.parse(
+        JSON.stringify(verifiablePresentationWithCredentialEIP712Types)
+      );
+
+    const { oneVCHasExpirationDate, allVCHaveExpirationDate } =
```
JGiter, line 316: This can shortened to two lines if using `some` and `every` on `verifiableCredential`

JGiter (approved):

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
Harasz: That’s true and with two loops instead of one as it is right now 😄

PR #556, opened 2022-05-16
fix: retry failed requests on token refresh

Harasz:
### Description

Fix previously failed request to SSI-Hub on token refresh.

### Contributor checklist

### DIFF
--- a/src/modules/cache-client/cache-client.service.ts
+++ b/src/modules/cache-client/cache-client.service.ts
@@ -21,6 +21,7 @@ import { cacheConfigs } from '../../config/cache.config';
 import { ICacheClient } from './cache-client.interface';
 import {
   AssetsFilter,
+  AuthTokens,
   ClaimsFilter,
   TEST_LOGIN_ENDPOINT,
 } from './cache-client.types';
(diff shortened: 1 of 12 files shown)


src/modules/cache-client/cache-client.service.ts
```diff
@@ -63,44 +64,39 @@ export class CacheClient implements ICacheClient {
...
   async authenticate() {
+    let tokens: AuthTokens | undefined = undefined;
+
+    // First try to refresh access token
     try {
       const refreshedTokens = await this.refreshToken();
 
-      if (refreshedTokens) {
-        getLogger().debug('[CACHE CLIENT] Setting authorization tokens');
-        if (!this.isBrowser) {
```
JGiter, line 71: Don't you think it will lead to failed authentication in line 73 in Node.js?

Harasz: I've tested it and everything is fine

src/modules/cache-client/cache-client.service.ts
```diff
@@ -132,7 +128,15 @@ export class CacheClient implements ICacheClient {
       getLogger().debug(`[CACHE CLIENT] axios error unauthorized`);
       const retryOriginalRequest = new Promise((resolve) => {
         this.failedRequests.push(() => {
-          resolve(axios(originalRequest));
+          console.log(originalRequest);
```
hejkerooo, line 131: remove console.log

hejkerooo (approved):

src/modules/signer/signer.service.ts
```diff
@@ -445,8 +445,10 @@ export class SignerService {
    *
```
JGiter, line 445:
```suggestion
   * @param force when true recalculates token even if it is already present
```

src/modules/cache-client/cache-client.service.ts
JGiter: Ok. Lets try it

JGiter (approved):

src/modules/cache-client/cache-client.service.ts
```diff
@@ -63,44 +64,43 @@ export class CacheClient implements ICacheClient {
...
         'Authorization'
-      ] = `Bearer ${token}`;
+      ] = `Bearer ${tokens.token}`;
     }
-    this.refresh_token = refreshToken;
-    this.pubKeyAndIdentityToken = pubKeyAndIdentityToken;
+    this.refresh_token = tokens.refreshToken;
 
+    // Run previously failed requests
+    console.log(
```
jrhender, line 101: Instead of `console.log` here, should it be replaced with `getLogger().log`?

jrhender (commented):
Seems like good changes to me, thanks @Harasz  We just need to be sure to test login/logout/change-account scenarios very thoroughly on Switchboard board.

Harasz:
@jrhender I've added jira task to add test for auth scenarios. We don't have one currently. And should be `getLogger`. But I will add task to use logger in other modules.

PR #558, opened 2022-05-17
Task/icl 287 pick changes

Harasz:
### Description

Remove VC storage, initialize VC service earlier, add IPFS CID checking before trying to download claims from IPFS.

Picked changes from https://github.com/energywebfoundation/iam-client-lib/pull/542

### Contributor checklist


src/modules/claims/claims.types.ts
```diff
@@ -54,6 +55,7 @@ export interface Claim {
   isRejected?: boolean;
   namespace: string;
   redirectUri?: string;
+  vp?: VerifiablePresentation;
```
jrhender, line 58: @Harasz why does the `Claim` have a `vp` property. That seems strange to me. I would think that it might have a `vc` maybe corresponding to the issued VC, but not a `vp`

jrhender (changes requested):
Looks good but could you update the initialization instructions in the README based on this change as well please? https://github.com/energywebfoundation/iam-client-lib#2-connect-to-the-cache-server

src/modules/claims/claims.types.ts
jrhender: Right, I remember now thanks!

JGiter (approved):

jrhender:
@Harasz does the ICL README still need to be updated with the change to the initialization?

Harasz:
@jrhender yeah I see now that README also have initialization guide. I will update it in next PR.

jrhender:
@Harasz Actually don't worry about it, as we will need to update in this PR anyways https://github.com/energywebfoundation/iam-client-lib/pull/560

Harasz:
@jrhender and also here #559

PR #559, opened 2022-05-18
feat: add claim revocation

Harasz:
### Description

On-chain claim revocation features.

### Contributor checklist


src/modules/claims/claims.service.ts
```diff
@@ -113,6 +123,30 @@ export class ClaimsService {
   async init() {
     const chainId = this._signerService.chainId;
     this._claimManager = chainConfigs()[chainId].claimManagerAddress;
+
+    const signer = this._signerService.signer;
+    const provider = signer.provider;
+    const publicKey = await this._signerService.publicKey();
+    let ewSigner: EwSigner;
```
JGiter, line 130: why not export it from did-registry.service?

src/modules/claims/claims.service.ts
```diff
@@ -843,6 +877,198 @@ export class ClaimsService {
...
+    return await this._claimRevocation.revokeClaim(namespace, subject, revoker);
+  }
+
+  /**
+   * Revoke On-Chain issued claims of the given namespace for multiple subjects. Namespace must be the same for all subjects.
+   *
+   * ```typescript
+   * claimsService.revokeMultipleClaim({
+   *     claims: [{
+   *         namespace: 'root.roles.energyweb.iam.ewc',
```
JGiter, line 918: Why not move namespace from array elements to root of claims type?

src/modules/claims/claims.service.ts
```diff
@@ -843,6 +877,198 @@ export class ClaimsService {
...
+    ];
+    let namespace = claims[0]?.namespace;
+
+    if (claimIds) {
+      const claimsDetails = await Promise.all([
+        ...claimIds.map(async (claimId) => {
+          return await this.getClaimById(claimId);
+        }),
+      ]);
+      claimsDetailsToRevoke = [
```
JGiter, line 958: Is this means that claims should not be provided along with claimId's? If so we need to mention this in description or better exclude such possibility at all

src/modules/claims/claims.service.ts
Harasz: The idea behind that is to make interface simple, so you only provide `claims` or `claimIds`.

src/modules/claims/claims.service.ts
Harasz: Yes, for all revocation method you can specify claim details or just simply pass claim id. And claim id will always override claim details. I've added comments to documentation and example uses

src/modules/claims/claims.service.ts
Harasz: Done

JGiter (approved):

JGiter (approved):

PR #570, opened 2022-05-30
Push stable version before revocation

jrhender:
### Description

I think it could make sense to push a "stable" version of ICL before we get the revocation code in. 

- [] Check if the version increment is correct. Is this indeed a breaking change?


jrhender:
@Harasz I'm looking at the [Changlog](https://github.com/energywebfoundation/iam-client-lib/pull/570/files#diff-2bfab3db5cc1baf4c6d3ff6b19901926e3bdf4411ec685dac973e5fcff1c723b) and it mentions a breaking change of "`VerifiableCredentialsService` is initialized earlier." . I'm wondering if this is legitimately a breaking change as the `master` branch doesn't have `VerifiableCredentialsService` as far as I can tell.

Harasz:
If vc service is not in master then it's not a breaking change

jrhender:
> If vc service is not in master then it's not a breaking change

Ok, thanks. I think it is too hard to revert the tagging as a breaking change though as we have tags and releases already for several `6.0` alpha versions so I propose we leave the major version change as is. It is major in the sense that we have significant new functionality!

Harasz:
Yeah, it was a breaking change for development branch. Maybe develop branch should not bump version to major :thinking:

Harasz:
@jrhender maybe we can close it, as revocation features are already here

jrhender:
Closing as we will push a stable version once revocation/verification features are stabilized

PR #579, opened 2022-06-13
feat: add credential revocation

Harasz:
### Description

Add credential revocation feature using StatusList2021.

### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/types/status-list.types.ts
+++ b/src/modules/verifiable-credentials/types/status-list.types.ts
@@ -1,0 +1,60 @@
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
+
+export interface StatusList2021UnsignedCredential {
+  '@context': Array<string | Record<string, unknown>>;
+  id: string;
+  type: string[];
+  issuer: string;
+  issuanceDate: string;
+  credentialSubject: {
+    id: string;
+    type: string;
+    statusPurpose: string;
+    encodedList: string;
+  };
+}
+
+export interface StatusList2021CredentialSubject {
+  [x: string]: unknown;
+  id: string;
+  type: string;
+  statusPurpose: string;
+  encodedList: string;
+}
+
+export type StatusList2021Credential =
+  VerifiableCredential<StatusList2021CredentialSubject>;
+
+export interface CredentialRevocationDetailsResult {
+  /* Revoker of the credential */
+  revoker: string;
+
+  /* Timestamp of revocation */
+  timestamp: number;
+}
+
+export const statusList2021CredentialEIP712Types = {
+  EIP712Domain: [],
+  VerifiableCredential: [
+    { name: '@context', type: 'string[]' },
+    { name: 'id', type: 'string' },
+    { name: 'type', type: 'string[]' },
+    { name: 'issuer', type: 'string' },
+    { name: 'issuanceDate', type: 'string' },
+    { name: 'credentialSubject', type: 'StatusList2021' },
+    { name: 'proof', type: 'Proof' },
+  ],
+  StatusList2021: [
+    { name: 'id', type: 'string' },
+    { name: 'type', type: 'string' },
+    { name: 'statusPurpose', type: 'string' },
+    { name: 'encodedList', type: 'string' },
+  ],
+  Proof: [
+    { name: '@context', type: 'string' },
+    { name: 'verificationMethod', type: 'string' },
+    { name: 'created', type: 'string' },
+    { name: 'proofPurpose', type: 'string' },
+    { name: 'type', type: 'string' },
+  ],
+};
--- a/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
+++ b/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
@@ -29,9 +29,13 @@ import {
   CreatePresentationParams,
   verifiableCredentialEIP712Types,
   verifiablePresentationWithCredentialEIP712Types,
+  StatusList2021Credential,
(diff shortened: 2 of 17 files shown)


src/modules/verifiable-credentials/types/status-list.types.ts
```diff
@@ -0,0 +1,57 @@
...
+  issuanceDate: string;
+  credentialSubject: {
+    id: string;
+    type: string;
+    statusPurpose: string;
+    encodedList: string;
+  };
+}
+
+export class StatusList2021Credential {
```
nichonien, line 18: @Harasz Should we move `StatusList2021Credential` to `ew-did-registry`, as it would be consumed in verification as well?

nichonien, line 18: @Harasz Should we move `StatusList2021Credential` to `ew-did-registry`, as it would be consumed in verification as well?

Harasz: @nichonien Yes, we should

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -436,7 +440,10 @@ export abstract class VerifiableCredentialsServiceBase {
    * @returns true if the proof is valid
    */
   public async verify<T extends ICredentialSubject>(
-    vp: VerifiablePresentation | VerifiableCredential<T>,
+    vp:
+      | VerifiablePresentation
+      | VerifiableCredential<T>
+      | StatusList2021Credential,
```
JGiter, line 446: Isn't StatusList2021Credential a specialization of VerifiableCredential for some T?

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -559,6 +580,134 @@ export abstract class VerifiableCredentialsServiceBase {
...
+        messageSchema: statusList2021CredentialEIP712Types,
+      },
+    };
+
+    const keyType = {
+      kty: 'EC',
+      crv: 'secp256k1',
+      alg: 'ES256K-R',
+      key_ops: ['signTypedData'],
+    };
```
jrhender, line 621: It seems like this object is duplicated from here: https://github.com/energywebfoundation/iam-client-lib/blob/8ddef78cd45a1911b308ef95d35663f6fc9c7aa1/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts#L268 ?
Perhaps it would be worth extracting it as a shared constant (or maybe function that returns this object, to avoid code changing the object and causing bugs elsewhere)

I think it would also be worth documenting where or how these values were selected as well please 🙏

jrhender (commented):
PR looks great so far 👍

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -559,6 +580,134 @@ export abstract class VerifiableCredentialsServiceBase {
...
+      signedCredential
+    ) as StatusList2021Credential;
+
+    return await this._cacheClient.persistCredentialStatusUpdate(
+      statusList2021Credential
+    );
+  }
+
+  /**
+   * Revoke given verifiable credential with StatusList2021.
```
JGiter, line 660:
```suggestion
   * Checks if credential is revoked.
```

src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -559,6 +580,134 @@ export abstract class VerifiableCredentialsServiceBase {
...
+   * await verifiableCredentialsService.isRevoked(credential);
+   * ```
+   *
+   * @param {VerifiableCredential<RoleCredentialSubject>} credential verifiable credential
+   * @return true if credential is revoked
+   */
+  public async isRevoked(
+    credential: VerifiableCredential<RoleCredentialSubject>
+  ): Promise<boolean> {
+    const statusListCredential =
```
JGiter, line 672: Is this method equivalent to return (await revocationDetails(credential)) !== null?

src/modules/verifiable-credentials/types/status-list.types.ts
```diff
@@ -0,0 +1,60 @@
...
+  id: string;
+  type: string[];
+  issuer: string;
+  issuanceDate: string;
+  credentialSubject: {
+    id: string;
+    type: string;
+    statusPurpose: string;
+    encodedList: string;
+  };
```
jrhender, line 14: Would this make sense?
```suggestion
  credentialSubject: StatusList2021CredentialSubject
```

jrhender (approved):
One suggestion but it looks great!

src/modules/verifiable-credentials/types/status-list.types.ts
Harasz: It would, if we move it to `@ew-did-registry/credentials-interface` along with other Status List Credential types

jrhender: Sorry, I'm not really following. Why would moving `StatusList2021UnsignedCredential` to `@ew-did-registry/credentials-interface` impact our ability to make the above change?

Harasz: We improve readability and we can re-use this interface in SSI-Hub for example by storing all related interfaces in one place

jrhender: Ok, no worries, it's a minor point, let's merge as is

Harasz: Okay

PR #583, opened 2022-06-21
feat: use claim default valid period from role definition

Harasz:
### Contributor checklist

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -4,6 +4,7 @@ import jsonwebtoken from 'jsonwebtoken';
 import { v4 } from 'uuid';
 import {
   IRoleDefinition,
+  IRoleDefinitionV2,
   PreconditionType,
 } from '@energyweb/credential-governance';
 import { ClaimRevocation } from '@energyweb/onchain-claims';
@@ -30,7 +31,7 @@ import {
   IClaimRejection,
   IClaimRequest,
   RegistrationTypes,
-  defaultClaimExpiry,
+  eternityTimestamp,
   erc712_type_hash,
   proof_type_hash,
   typedMsgPrefix,
@@ -398,16 +399,13 @@ export class ClaimsService {
     const strippedClaimData = this.stripClaimData(claimData);
     const { claimType: role, claimTypeVersion: version } = claimData;
 
-    const defaultValidityPeriod = roleDefinition?.['defaultValidityPeriod'];
-    const roleDefinitionExpirationTimestamp = defaultValidityPeriod
-      ? Date.now() + defaultValidityPeriod
-      : undefined;
-
-    const claimExpirationTimestamp =
-      expirationTimestamp || roleDefinitionExpirationTimestamp;
+    const claimExpirationTimestamp = this.getClaimExpirationTimestamp(
+      roleDefinition as IRoleDefinitionV2,
+      expirationTimestamp
+    );
 
     if (registrationTypes.includes(RegistrationTypes.OnChain)) {
-      const expiry = claimExpirationTimestamp || defaultClaimExpiry;
+      const expiry = claimExpirationTimestamp || eternityTimestamp;
       const onChainProof = await this.createOnChainProof(
         role,
         version,
@@ -509,7 +507,7 @@ export class ClaimsService {
       onChainProof,
       expirationTimestamp,
     } = claim;
-    const expiry = expirationTimestamp || defaultClaimExpiry;
+    const expiry = expirationTimestamp || eternityTimestamp;
     const data = this._claimManagerInterface.encodeFunctionData('register', [
       addressOf(subject),
       namehash(claimType),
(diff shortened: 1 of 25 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -392,8 +398,16 @@ export class ClaimsService {
...
+    const roleDefinitionExpirationTimestamp = defaultValidityPeriod
+      ? Date.now() + defaultValidityPeriod
+      : undefined;
+
+    const claimExpirationTimestamp =
+      expirationTimestamp || roleDefinitionExpirationTimestamp;
+
     if (registrationTypes.includes(RegistrationTypes.OnChain)) {
-      const expiry = expirationTimestamp || defaultClaimExpiry;
+      const expiry = claimExpirationTimestamp || defaultClaimExpiry;
```
JGiter, line 410: Not related to this PR, but, I wonder is it worth replace `defaultClaimExpiry` by `Erternity` ?

src/modules/claims/claims.service.ts
```diff
@@ -582,14 +596,33 @@ export class ClaimsService {
...
 
     const message: IClaimIssuance = {
       id: v4(),
       requester: subject,
       claimIssuer: [this._signerService.did],
       acceptedBy: this._signerService.did,
     };
+
+    const defaultValidityPeriod = roleDefinition?.['defaultValidityPeriod'];
+    const roleDefinitionExpirationTimestamp = defaultValidityPeriod
```
JGiter, line 619: Looks like this variable doesn't really needed. `claimExpirationTimestamp` can be calculated by this formula by replacing undefined to expirationTimestamp || defaultExpirationTimestamp

JGiter (approved):

src/modules/claims/claims.service.ts
JGiter: It just a new name. To express expiry which never expires

Harasz: Ohh okay, make sense.

src/modules/claims/claims.service.ts
Harasz: Yeah I know, I make for more clarity in code

src/modules/claims/claims.service.ts
nichonien: @JGiter @Harasz We can remove the `defaultClaimExpiry`. IMO either the expiry should come from the `RoleDefinition` or the `issuer` itself. If let's say the issuer wants it to be non-expiry claim, then they themselves should pass this a value.

Harasz: @jrhender what do you think about it?

src/modules/claims/claims.service.ts
```diff
@@ -392,8 +398,16 @@ export class ClaimsService {
     const strippedClaimData = this.stripClaimData(claimData);
     const { claimType: role, claimTypeVersion: version } = claimData;
 
+    const defaultValidityPeriod = roleDefinition?.['defaultValidityPeriod'];
+    const roleDefinitionExpirationTimestamp = defaultValidityPeriod
+      ? Date.now() + defaultValidityPeriod
+      : undefined;
```
jrhender, line 404: I think there is some duplication here and the same snippet in `issueClaim`. Perhaps it makes sense to extract into a private method?

src/modules/claims/claims.service.ts
```diff
@@ -599,15 +632,15 @@ export class ClaimsService {
       const [issuedToken, vp] = await Promise.all([
         this._didRegistry.issuePublicClaim({
           publicClaim,
-          expirationTimestamp,
+          expirationTimestamp: claimExpirationTimestamp,
```
jrhender, line 635: Are you sure it makes sense to pass `expirationTimestamp` even if `claimExpirationTimesamp === undefined`. To me that is relying a bit too much on the implementation of `issuePublicClaim` to not add an `exp` to the JWT if an `undefined` value is passed (which I think is the behaviour we would want `issuePublicClaim` to have)

src/modules/claims/claims.service.ts
jrhender: I'm open to the rename to `eternity`

Regarding removing the `defaultClaimExpiry`, I think we need it because the contract requires expiry https://github.com/energywebfoundation/ew-credentials/blob/c54672118a2687146e8421d3779208a04ee191c0/packages/onchain-claims/contracts/ClaimManager.sol#L93

src/modules/claims/claims.service.ts
```diff
@@ -599,15 +632,15 @@ export class ClaimsService {
...
         }),
         this.issueVerifiablePresentation({
           subject,
           namespace: claim.claimType,
           version: claim.claimTypeVersion.toString(),
           issuerFields: claim.issuerFields,
           credentialStatus,
-          expirationTimestamp,
+          expirationTimestamp: claimExpirationTimestamp,
```
jrhender, line 643: Similar to other comment, we can be sure that now (and in the future) `issueVerifiablePresentation` won't add an `expiry` prop to the VC if this `expirationTimestamp` is `undefined`?

Harasz: I am sure that it won't add this props unless it has some value other the undefined and null or even 0

src/modules/claims/claims.service.ts
jrhender: @Harasz Are you also sure in this case that the expiration wouldn't be added to the JWT if it is undefined?

Harasz: Yes, i am sure. For javascript both cases are the same. lets do simple test
```
const a = {prop1: undefined}
console.log(a.prop1) // undefined
console.log(a.prop2) // no, error just unefined
console.log(a) // { prop1: undefined }
```
The only difference is that third console log show prop1

jrhender: Ok, I was concerned that this could result in JWTs as shown below (as this would be an [invalid JWT](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4), I believe) 
```json
{
  "exp": undefined
}
```
But I realized now that this isn't even valid JSON, so think that this pretty unlikely to happen

jrhender (approved):

JGiter (approved):

PR #584, opened 2022-06-22
fix: cache-client token refresh

Harasz:
### Description

Fix token refreshing

### Contributor checklist

### DIFF
--- a/src/modules/cache-client/cache-client.service.ts
+++ b/src/modules/cache-client/cache-client.service.ts
@@ -76,12 +76,24 @@ export class CacheClient implements ICacheClient {
   async authenticate() {
     let tokens: AuthTokens | undefined = undefined;
 
+    const setTokens = () => {
+      if (!tokens) return;
+      if (!this.isBrowser) {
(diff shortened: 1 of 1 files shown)


src/modules/cache-client/cache-client.service.ts
```diff
@@ -76,12 +76,24 @@ export class CacheClient implements ICacheClient {
...
+
     // First try to refresh access token
     try {
       const refreshedTokens = await this.refreshToken();
+      tokens = refreshedTokens;
+      setTokens();
 
-      if (refreshedTokens && (await this.isAuthenticated())) {
-        tokens = refreshedTokens;
+      if (!(await this.isAuthenticated())) {
```
JGiter, line 95: We probably should also find out how authentication may fail right after refreshing token

JGiter (approved):

src/modules/cache-client/cache-client.service.ts
Harasz: it shouldn't but you know, better to check it :smile:

PR #586, opened 2022-06-23
feat(RevokeClaim): add endpoint to fetch claims by revoker (current u…

whitneypurdum:
### Description
- Add cache client endpoint to fetch claims by revoker (current user)
- Add claims service method to call cache client

### Contributor checklist


JGiter (approved):
Looks good. We can also set default revoker, as this is will be common case

whitneypurdum:
> Looks good. We can also set default revoker, as this is will be common case

Thanks @JGiter !

PR #595, opened 2022-07-07
task/ICL-323 high resource usage of importing iam-lib

JGiter:
### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/verifiable-credentials-node.service.ts
+++ b/src/modules/verifiable-credentials/verifiable-credentials-node.service.ts
@@ -1,60 +1,78 @@
-import {
-  prepareIssueCredential,
-  completeIssueCredential,
-  verifyCredential,
-  verifyPresentation,
-} from 'didkit-wasm-node';
-import {
-  prepareIssuePresentation,
-  completeIssuePresentation,
-} from '@spruceid/didkit-wasm-node';
-import { SignerService } from '../signer';
+import { SignerService } from '../signer/signer.service';
+import { CacheClient } from '../cache-client';
 import { VerifiableCredentialsServiceBase } from './verifiable-credentials-base.service';
 
 export class VerifiableCredentialsServiceNode extends VerifiableCredentialsServiceBase {
-  protected prepareIssueCredential: (
-    credential: string,
-    linked_data_proof_options: string,
-    public_key: string
-  ) => Promise<string>;
-  protected completeIssueCredential: (
-    credential: string,
-    preparation: string,
-    signature: string
-  ) => Promise<string>;
-  protected verifyCredential: (
-    vc: string,
-    proof_options: string
-  ) => Promise<string>;
 
-  protected prepareIssuePresentation: (
-    presentation: string,
-    linked_data_proof_options: string,
-    public_key: string
-  ) => Promise<string>;
-  protected completeIssuePresentation: (
-    presentation: string,
-    preparation: string,
-    signature: string
-  ) => Promise<string>;
-  protected verifyPresentation: (
-    vp: string,
-    proof_options: string
-  ) => Promise<string>;
-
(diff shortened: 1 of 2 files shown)


JGiter:
> I think that this fix will only fix an error during initialization. But problem will still remain when will try to use VC features, so during issuing claim request and spike will be always when invoking one of the VC features

From my opinion high usage was caused by blocking cpu on reading from disk. Now reading is asynchronous and Node.js will be able to yield control from io operation

JGiter:
> Issue still occurs, but i think it's not that big at it was before: 160MB and 400%CPU, previously it was 800MB. Hovewever issue occurs every time you use one of the features of VC module @JGiter @jrhender

You are right. Issue still remains, though there is some difference. Here test of verifiable credentials with static and dynamic import
image
image

jrhender:
@JGiter I'm confused by what your graph is comparing? Is the top the behaviour without this PR and the bottom is the behaviour with the PR?

You mentioned "static and dynamic import", but I believe both with this PR and before this PR, "dynamic" imports are being used. "Dynamic import" just refers to the use of `import()` at runtime, right?

JGiter:
> @JGiter I'm confused by what your graph is comparing? Is the top the behaviour without this PR and the bottom is the behaviour with the PR?
The top graph is testing with dynamic import, the bottom - with static

> You mentioned "static and dynamic import", but I believe both with this PR and before this PR, "dynamic" imports are being used. "Dynamic import" just refers to the use of `import()` at runtime, right?
Now we are statically importing members from didkit, for example here https://github.com/energywebfoundation/iam-client-lib/blob/9fd459e9d3b8d092880adc17065e2651d0dca812/src/modules/verifiable-credentials/verifiable-credentials-node.service.ts#L6

jrhender:
@JGiter 
> You mentioned "static and dynamic import", but I believe both with this PR and before this PR, "dynamic" imports are being used. "Dynamic import" just refers to the use of `import()` at runtime, right?

Could you clarify which code each of your graphs was executing? Is one graph the current `develop` branch and one graph is this PR branch? Or if it is some different code that was executed, could you point to that code please?

JGiter:
@jrhender 
> Could you clarify which code each of your graphs was executing? Is one graph the current `develop` branch and one graph is this PR branch? Or if it is some different code that was executed, could you point to that code please?

Top graph is result of execution of code in this PR, the bottom graph corresponds to develop

jrhender:
@JGiter 

> Top graph is result of execution of code in this PR, the bottom graph corresponds to develop

Ok, so I think in both cases "dynamic" imports were being used, right?

In this PR branch: https://github.com/energywebfoundation/iam-client-lib/blob/bf8e6edf8cb066f09e44be0cd0fdb05cf76b6788/src/modules/verifiable-credentials/index.ts#L15 and https://github.com/energywebfoundation/iam-client-lib/blob/bf8e6edf8cb066f09e44be0cd0fdb05cf76b6788/src/modules/verifiable-credentials/verifiable-credentials-node.service.ts#L16

In develop branch: https://github.com/energywebfoundation/iam-client-lib/blob/fcbd34e8abe6fb79ae440b5a5c3646ae4731aa5d/src/modules/verifiable-credentials/index.ts#L15

So I am confused by what you mean by the top graph being "with static imports". If it was using the code from this PR, then it is still using dynamic imports, right?

JGiter:
@jrhender 
> So I am confused by what you mean by the top graph being "with static imports". If it was using the code from this PR, then it is still using dynamic imports, right?

Ah, I see. Indeed, verifiable-credential is imported dynamically in both branches. But verifiable-credential doesn't contain resource intensive operations, so I think the way of importing of vc is not important. This PR is differs from develop by importing didkit. This import is static in develop

JGiter:
@jrhender Should we close this?

jrhender:
@JGiter yes, I think so, thanks

PR #596, opened 2022-07-07
test(cache-client): add authorization tests

Harasz:
### Contributor checklist

### DIFF
--- a/src/modules/cache-client/cache-client.service.ts
+++ b/src/modules/cache-client/cache-client.service.ts
@@ -583,11 +583,11 @@ export class CacheClient implements ICacheClient {
   }
 
   /**
-   * @description Interceptor of authentication errors. Queues failed requests and starts authentication process.
+   * Decides whether to retry the request or not based on the given axios error.
    *
-   * @param error Intercepted response from failed request
+   * @param error axios error
    *
-   * @returns Promise, which resolves with result of resending of failed request
+   * @return true if request should be retried
    */
   private async handleRequestError(error: Error): Promise<boolean> {
     if (!axios.isAxiosError(error)) {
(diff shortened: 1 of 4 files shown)


src/modules/cache-client/cache-client.service.ts
```diff
@@ -598,20 +602,37 @@ export class CacheClient implements ICacheClient {
...
+      return false;
+    }
+
+    if (!isAuthError) {
+      return true;
+    }
+
     getLogger().debug(`[CACHE CLIENT] axios error unauthorized`);
 
     if (!this.isAuthenticating) {
```
JGiter, line 638: Would it make sense to put isAuthenticating = true and isAuthenticating = false to authenticate()?

src/modules/cache-client/cache-client.service.ts
```diff
@@ -598,20 +602,37 @@ export class CacheClient implements ICacheClient {
...
+    // Retry some client errors
+    if (
+      response.status >= 400 &&
+      response.status < 500 &&
+      !clientErrorsToRetry.includes(response.status)
+    ) {
+      return false;
+    }
+
+    if (!isAuthError) {
```
JGiter, line 632: Why not place it to where it is defined?

src/modules/cache-client/cache-client.service.ts
```diff
@@ -598,20 +602,37 @@ export class CacheClient implements ICacheClient {
...
+        config.url.indexOf('/refresh_token') >= 0 ||
+        config.url.indexOf(TEST_LOGIN_ENDPOINT) >= 0);
 
-    if (!isAuthError || isAuthEndpoint) {
+    if (isAuthEndpoint) {
       return false;
     }
 
+    // Retry some client errors
+    if (
```
JGiter, line 624: Is this the same as `if (clientErrorsToRetry.includes(...`? Also we can check this without introducing new var similarly as isAuthError

JGiter (commented):
Description of handleRequestError is not up-to-date. Taking chance we can fix this as well as method name

src/modules/cache-client/cache-client.service.ts
Harasz: We would need to wrap the whole function in try-catch, now its only two lines. I was thinking about it, but imo current approach is more readable

src/modules/cache-client/cache-client.service.ts
Harasz: moved

src/modules/cache-client/cache-client.service.ts
Harasz: I know, but i think this is more readable, if you read the var name and you know what is going on instead of  analyzing the code

JGiter (approved):

PR #598, opened 2022-07-13
benchmarking

JGiter:
Benchmark initialization

### Contributor checklist


.gitignore
```diff
@@ -12,3 +12,5 @@ compiled
 ethers
 site
 ethers-staking
+*.heapprofile
+*.cpuprofile
```
jrhender, line 16: Are these files being created by `runner.js` I didn't notice that this was the case.

benchmarks/runner.js
```diff
@@ -0,0 +1,41 @@
+const { execFile } = require('child_process');
```
jrhender, line 1:
```suggestion
/*
 * This script needs to be run on linux, due to use of `top` command
 */

const { execFile } = require('child_process');
```

jrhender (approved):
I think we can merge it in. Seems like a useful benchmarking script to have around.

.gitignore
JGiter: They are created by Node.js profiler. We don't need them now. I will remove them

PR #600, opened 2022-07-14
task/ICL-311 verify credential status

JGiter:
### Contributor checklist


e2e/credential-revocation.e2e.ts
```diff
@@ -277,13 +294,21 @@ describe('Off-chain credential revocation', () => {
 
     const verifiableCredential = extractCredential(claim);
 
-    mockCredentialInitiate(issuer.signerService.didHex);
+    mockCredentialInitiate(revoker.signerService.didHex);
```
jrhender, line 297: Could this be a bit more accurate? It is initiating the revocation?
```suggestion
    mockCredentialInitiateRevocation(revoker.signerService.didHex);
```

jrhender (commented):
Looks like a good start, thanks @JGiter

e2e/claims.service.e2e.ts
```diff
@@ -826,7 +827,7 @@ describe('Сlaim tests', () => {
       });
     };
 
-    test('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
+    test.only('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
```
jrhender, line 830: Probably we don't want the `only` here, right?
```suggestion
    test('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
```

e2e/credential-revocation.e2e.ts
```diff
@@ -310,15 +335,25 @@ describe('Off-chain credential revocation', () => {
...
+    expect(statusListCredential).toBeDefined();
+    nock(verifiableCredential.credentialStatus?.statusListCredential as string)
+      .get('')
+      .reply(200, statusListCredential);
+
+    expect(await issuer.claimsService.verifyVc(verifiableCredential)).toEqual(
+      expect.objectContaining({
+        isVerified: false,
+        errors: expect.arrayContaining([
+          expect.stringContaining('StatusList credential is invalid'), // part of error message of InvalidStatusList
```
jrhender, line 353: Should we export error messages from ew-credentials to avoid needing to hardcode this string? (would be a separate story)

jrhender (approved):
A small question, but it looks good to me!

e2e/claims.service.e2e.ts
JGiter: oh, sorry. How do think should we add this https://www.npmjs.com/package/eslint-plugin-no-only-tests?

e2e/credential-revocation.e2e.ts
JGiter: We can get error message from error classes which are already exported. This string appeared because message of https://github.com/energywebfoundation/ew-did-registry/blob/403b8dc790d2abcd36e9ddb56c6e1afca43f12a8/packages/revocation/src/errors.ts#L1 is constructed from more specific messages received from `verify`, which we don't know beforehand. But now I realized it can be achieved as `new InvalidStatusList(['']).message`

jrhender (approved):

PR #601, opened 2022-07-14
test(verify): Clean up verify tests

whitneypurdum:
### Description

- Remove unneeded role creation
- Wrap verify tests in describe
- Add test for no resolved credential

### Contributor checklist

### DIFF
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -826,17 +826,12 @@ describe('Сlaim tests', () => {
         expirationDate,
       });
     };
-    describe('Verify off-chain claim and credential', () => {
+    describe('Verify off-chain claim (EIP191JWT) and credential', () => {
       test('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
         await signerService.connect(rootOwner, ProviderType.PrivateKey);
-        await domainsService.createRole({
-          roleName: verifyVcRole,
-          namespace,
-          data: roles[`${verifyVcRole}.${root}`],
-          returnSteps: false,
-        });
+        const issuerFields = [];
         const vc = await createExampleSignedCredential(
-          [],
+          issuerFields,
           `${verifyVcRole}.${root}`
         );
         nock(vc.credentialStatus?.statusListCredential as string)
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -826,112 +826,90 @@ describe('Сlaim tests', () => {
         expirationDate,
       });
     };
-
-    test('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
-      await signerService.connect(rootOwner, ProviderType.PrivateKey);
-      await domainsService.createRole({
-        roleName: verifyVcRole,
-        namespace,
-        data: roles[`${verifyVcRole}.${root}`],
-        returnSteps: false,
+    describe('Verify off-chain claim and credential', () => {
+      test('verifyVc should verify a VC with no errors if the issuer is authorized', async () => {
+        await signerService.connect(rootOwner, ProviderType.PrivateKey);
+        await domainsService.createRole({
+          roleName: verifyVcRole,
+          namespace,
+          data: roles[`${verifyVcRole}.${root}`],
+          returnSteps: false,
+        });
+        const vc = await createExampleSignedCredential(
+          [],
+          `${verifyVcRole}.${root}`
+        );
+        nock(vc.credentialStatus?.statusListCredential as string)
+          .get('')
+          .reply(200, undefined);
+        const result = await claimsService.verifyVc(vc);
+        expect(result.errors).toHaveLength(0);
+        expect(result.isVerified).toBe(true);
       });
-      const vc = await createExampleSignedCredential(
-        [],
-        `${verifyVcRole}.${root}`
-      );
-      nock(vc.credentialStatus?.statusListCredential as string)
-        .get('')
-        .reply(200, undefined);
-      const result = await claimsService.verifyVc(vc);
-      expect(result.errors).toHaveLength(0);
-      expect(result.isVerified).toBeTruthy;
-    });
-    test('resolveCredentialAndVerify should resolve a Verifiable Credential and call correct verification method (verifyVC)', async () => {
-      const roleName = `${resolveVC}.${root}`;
-      await signerService.connect(rootOwner, ProviderType.PrivateKey);
-      await domainsService.createRole({
-        roleName: resolveVC,
-        namespace,
-        data: roles[roleName],
-        returnSteps: false,
+      test('resolveCredentialAndVerify should resolve and verify a Verifiable Credential', async () => {
+        const roleName = `${resolveVC}.${root}`;
+        const { issuedToken } = await enrolAndIssue(rootOwner, staticIssuer, {
+          subjectDID: rootOwnerDID,
+          claimType: roleName,
+          registrationTypes: [
+            RegistrationTypes.OnChain,
+            RegistrationTypes.OffChain,
+          ],
+          publishOnChain: true,
+          issuerFields: [],
+        });
+        await signerService.connect(rootOwner, ProviderType.PrivateKey);
+        const subjectDoc = await didRegistry.getDidDocument({
+          did: rootOwnerDID,
+          includeClaims: true,
+        });
+        mockCachedDocument.mockResolvedValueOnce(subjectDoc);
+        await claimsService.publishPublicClaim({
+          claim: { token: issuedToken },
+        });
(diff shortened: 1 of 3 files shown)


e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+          subjectDID: rootOwnerDID,
+          claimType: roleName,
+          registrationTypes: [
+            RegistrationTypes.OnChain,
+            RegistrationTypes.OffChain, // role type issuer should have offchain claim
+          ],
+          publishOnChain: true,
+          issuerFields: [],
+        });
+        await signerService.connect(rootOwner, ProviderType.PrivateKey);
```
JGiter, line 854: I think there is no need to change signer to get document from chain
```suggestion
        
```

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+        const roleName = `${resolveVC}.${root}`;
+        await signerService.connect(rootOwner, ProviderType.PrivateKey);
+        await signerService.connect(staticIssuer, ProviderType.PrivateKey);
+        await createExampleSignedCredential([], roleName);
+        const { issuedToken } = await enrolAndIssue(rootOwner, staticIssuer, {
+          subjectDID: rootOwnerDID,
+          claimType: roleName,
+          registrationTypes: [
+            RegistrationTypes.OnChain,
+            RegistrationTypes.OffChain, // role type issuer should have offchain claim
```
JGiter, line 849: This comment isn't needed here as root owner is not issue
```suggestion
            RegistrationTypes.OffChain,
```

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
-        publishOnChain: true,
-        issuerFields: [],
-      });
-      await signerService.connect(rootOwner, ProviderType.PrivateKey);
-      let subjectDoc = await didRegistry.getDidDocument({
-        did: rootOwnerDID,
-        includeClaims: true,
-      });
-      mockCachedDocument.mockResolvedValueOnce(subjectDoc);
+        subjectDoc = await didRegistry.getDidDocument({
```
JGiter, line 865: Do you want to check that claim was published?

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+          rootOwnerDID,
+          roleName
+        );
+        expect(result.errors).toHaveLength(0);
+        expect(result.isVerified).toBeTruthy;
+        expect(claimsService.verifyVc).toHaveBeenCalled;
+      });
+      test('resolveAndVerify should resolve an EIP191JWT and verify the claim', async () => {
+        const roleName = `${verifyOffChainClaimRole}.${root}`;
+        await signerService.connect(staticIssuer, ProviderType.PrivateKey);
```
JGiter, line 879: Are you connecting to static issuer to issue? Issuer is connected in `enrolAndIssue`

JGiter (changes requested):

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
-        claim: { token: issuedToken },
-      });
+        await claimsService.publishPublicClaim({
+          claim: { token: issuedToken },
+        });
 
-      subjectDoc = await didRegistry.getDidDocument({
-        did: rootOwnerDID,
-        includeClaims: true,
+        subjectDoc = await didRegistry.getDidDocument({
```
jrhender, line 901: It seems `subjectDoc` isn't used?

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+          did: rootOwnerDID,
+          includeClaims: true,
+        });
+        const result = await claimsService.resolveCredentialAndVerify(
+          rootOwnerDID,
+          roleName
+        );
+        expect(result.errors).toHaveLength(0);
+        expect(result.isVerified).toBeTruthy;
+        expect(claimsService.verifyRoleEIP191JWT).toHaveBeenCalled;
```
jrhender, line 911: IMO, it is typically best not to assert function/method calls as these are implementation details. Is it necessary to do so in this case?

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+        );
+        const result = await claimsService.verifyVc(vc);
+        expect(result.errors).toHaveLength(0);
+        expect(result.isVerified).toBeTruthy;
       });
+      test('resolveCredentialAndVerify should resolve a Verifiable Credential and call correct verification method (verifyVC)', async () => {
+        const roleName = `${resolveVC}.${root}`;
+        await signerService.connect(rootOwner, ProviderType.PrivateKey);
+        await signerService.connect(staticIssuer, ProviderType.PrivateKey);
+        await createExampleSignedCredential([], roleName);
```
jrhender, line 843: Is this needed? It seems the return result isn't being assigned to a variable.

Perhaps there is a side-effect, however if that is the case, I think it is better for functions with side-effects to have a `void` return as this suggests to the reader that the function performs a side-effect (as otherwise, the void-return function wouldn't be useful).
Maybe this is old hat to you, but in other words, in clean code, IMO, there should be two types of functions:
- `pureFunction(input1, input2): returnType`
- `functionWithSideEffects(input1): void`

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
+          did: rootOwnerDID,
+          includeClaims: true,
+        });
+        const result = await claimsService.resolveCredentialAndVerify(
+          rootOwnerDID,
+          roleName
+        );
+        expect(result.errors).toHaveLength(0);
+        expect(result.isVerified).toBeTruthy;
+        expect(claimsService.verifyVc).toHaveBeenCalled;
```
jrhender, line 875: Best not to test implementation details, IMO. (Unless there is a good reason in this case)

jrhender (commented):
Overall seems good, just some comments for discussion

e2e/claims.service.e2e.ts
jrhender: Yes, I mean the use of `toHaveBeenCalled`. In my opinion, rather than testing if a specific function has been called, I think that it is better to have tests that confirm the effect of the function. The way to do that, in my opinion, is to test before the positive and negative cases. So this test is `resolveCredentialAndVerify should resolve a Verifiable Credential and call correct verification method (verifyVC)`, so I would change to:
- `resolveCredentialAndVerify should resolve and verify a Verifiable Credential` -> a rename of this test
- `resolveCredentialAndVerify should throw error if <some case that verfiyVC covers>` (maybe you have made this test already)

jrhender (commented):
Some responses

e2e/claims.service.e2e.ts
whitneypurdum: It's been removed

e2e/claims.service.e2e.ts
```diff
@@ -825,109 +825,103 @@ describe('Сlaim tests', () => {
...
-      await signerService.connect(rootOwner, ProviderType.PrivateKey);
-      let subjectDoc = await didRegistry.getDidDocument({
-        did: rootOwnerDID,
-        includeClaims: true,
-      });
-      mockCachedDocument.mockResolvedValueOnce(subjectDoc);
-
-      await claimsService.publishPublicClaim({
-        claim: { token: issuedToken },
+    describe('Verify off-chain claim and credential', () => {
```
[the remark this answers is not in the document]
whitneypurdum: ALso included in this describe block is verifyVC. I can make another describe block for resolveAndVerify if we want that and make top-level remain as Verify off-chain claim and credential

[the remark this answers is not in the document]
jrhender: > I can make another describe block for resolveAndVerify if we want that and make top-level remain as Verify off-chain claim and credential

This sounds good to me!

JGiter (approved):

src/modules/claims/claims.service.ts
```diff
@@ -479,6 +479,7 @@ export class ClaimsService {
           ...strippedClaimData,
           ...(issuerFields && { issuerFields }),
         },
+        //credentialStatus: credentialStatus
```
jrhender, line 482:
```suggestion
```

jrhender (approved):
It's looking good to me, just a suggestion to remove a commented line

PR #602, opened 2022-07-18
refactor: cleanup credential exchange code

Harasz:
### Description

Validate exchange credential types

### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
+++ b/src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
@@ -4,6 +4,7 @@ import {
   PEX,
   SelectResults,
 } from '@sphereon/pex';
+import { InputDescriptorV1, InputDescriptorV2 } from '@sphereon/pex-models';
 import { v4 as uuid } from 'uuid';
 import axios from 'axios';
 import {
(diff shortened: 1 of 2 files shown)


src/modules/verifiable-credentials/verifiable-credentials-base.service.ts
```diff
@@ -702,4 +697,22 @@ export abstract class VerifiableCredentialsServiceBase {
...
+   * This could be related to these issues:
+   *    - https://github.com/Sphereon-Opensource/pex/issues/96
+   *    - https://github.com/Sphereon-Opensource/pex/issues/91
+   * We are therefore trying to simplify the input to pex so remove the possibility of it generating incorrect results.
+   * Once the above issues are fixed, pex can be updated and perhaps this filtering will not needed.
+   *
+   * @param {InputDescriptorV1 | InputDescriptorV2} descriptors input descriptors
+   * @returns filtered input descriptors
+   */
+  private filterSelfSignDescriptors(
```
jrhender, line 713: @Harasz I guess your view is that `private` methods should be at the end of file? Is that what is motivating this change?

Harasz: yes, we are following that rule in other modules

jrhender: I agree with this convention. For me, this convention is coming from my experience with the C# ecosystem https://stackoverflow.com/a/310967 . Do you know where you got this convention from?

Harasz: you forced it somewhere and i liked this approach so you know :smile:

jrhender (approved):
Looks good to me

PR #606, opened 2022-07-18
feat(revocation): add credentialStatus to Public Claim

whitneypurdum:
### Description

- add credentialStatus to publicClaim
- if the credentialStatusOverride value is passed in, prefer it to the vp's credentialStatus
### Contributor checklist


whitneypurdum:
Looking at how tests are written, I don't think that this needs an additional test as we were already making the vp call and we are not checking the keys of the publicClaim, but I could be wrong about this...

src/modules/claims/claims.service.ts
```diff
@@ -472,30 +473,30 @@ export class ClaimsService {
...
+        version: version.toString(),
+        issuerFields,
+        credentialStatus: credentialStatusOverride,
+        expirationTimestamp: claimExpirationTimestamp,
+      });
+      const credStatus = vp?.verifiableCredential[0]?.credentialStatus;
       const publicClaim: IPublicClaim = {
         did: sub,
         signer: this._signerService.did,
+        credentialStatus: credentialStatusOverride || credStatus,
```
jrhender, line 488: I'm not sure we want to add the `credentialStatus` property if both `credentialStatusOverride || credStatus` are undefined/null. I think it would be better in this case if `credentialStatus` wasn't specified? 🤔 

I used this script to test that JS still adds props to objects if the value is undefined. https://playcode.io/new/
```
let i = undefined;
const message = {
  prop: i
}

// Log to console
console.log(message)
console.log('prop' in message)
console.log(message.hasOwnProperty('prop'))
```

jrhender:
@whitneypurdum 
>Looking at how tests are written, I don't think that this needs an additional test as we were already making the vp call and we are not checking the keys of the publicClaim,

Would it be possible to test that the output EIP191JWT has the `credentialStatus` property?

src/modules/claims/claims.service.ts
```diff
@@ -663,28 +667,28 @@ export class ClaimsService {
     };
 
     if (registrationTypes.includes(RegistrationTypes.OffChain)) {
+      const vp = await this.issueVerifiablePresentation({
```
whitneypurdum, line 670: This code is for this task. The other changes in this file were changes needed after updating ew-credentials version

jrhender: Can you rebase against `develop`  please? I think https://github.com/energywebfoundation/iam-client-lib/pull/615 updated ew-credential as well (to an even higher version).
Then I think this PR can be isolate to just "add credentialStatus to PublicClaim"

whitneypurdum: just rebased this! thanks

jrhender (approved):
Looks good to me, thanks @whitneypurdum

PR #607, opened 2022-07-18
chore(logger): replace console log with logger

whitneypurdum:
### Description

Remove console.logs and replace with logger

### Contributor checklist


whitneypurdum:
This was the only module that had console logging. Other instances of console logs were in the /scripts directory - I had originally changed to use logger there but then realized the task specifically stated it should be changed for modules.

Harasz:
it's not only replacing the console log to logger, but also adding logger in other places. So we can easy debug the any problems in the future

jrhender (approved):
I think this is an improvement as is, so I think it worth it to merge as is

@whitneypurdum Just a reminder to either "Squash merge" the PR via the Github UI or `rebase -i` squash the commits before merging please

jrhender:
@Harasz I think we can add other tasks/story to add additional logs

PR #608, opened 2022-07-19
task/ICL-314 can't issue without issuer fields value

JGiter:
### Contributor checklist

### DIFF
--- a/src/modules/verifiable-credentials/types/verifiable-credentials.types.ts
+++ b/src/modules/verifiable-credentials/types/verifiable-credentials.types.ts
@@ -1,6 +1,7 @@
+import { IssuerFields } from '@energyweb/credential-governance';
 import { StatusList2021Entry } from '@ew-did-registry/credentials-interface';
 import { IPresentationDefinition, SelectResults } from '@sphereon/pex';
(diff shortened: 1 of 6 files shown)


src/modules/verifiable-credentials/types/verifiable-credentials.types.ts
```diff
@@ -28,6 +29,19 @@ export interface RoleCredentialSubjectParams {
   /** Indicates if credential is actual of time of verification */
   credentialStatus?: StatusList2021Entry;
 }
+export const validateRoleCredentialSubject = (
+  subject: RoleCredentialSubjectParams
+) => {
+  const field = subject.issuerFields?.find(
+    (field) => !['string', 'number'].includes(typeof field.value)
+  );
+  if (field) {
```
jrhender, line 38: I think this is a bit easier to read
```suggestion
  const invalidField = subject.issuerFields?.find(
    (field) => !['string', 'number'].includes(typeof field.value)
  );
  if (invalidField) {
```

jrhender (approved):
One suggestion but seems alright to me

Harasz (approved):

PR #612, opened 2022-07-21
task/ICS-188 incorrect claim status

JGiter:
### Description

We need to make sure that failed registering on-chain doesn't prevent posting of issued claim. Registration onchain fails for two reasons: issuer isn't authorized or role already registered. First of this is checked now. This PR adds another check

### Contributor checklist


src/modules/claims/claims.service.ts
```diff
@@ -445,29 +443,30 @@ export class ClaimsService {
...
-          expirationTimestamp: expiry,
-        });
-      }
+      await this.registerOnchain({
+        token,
+        subjectAgreement,
+        onChainProof,
+        acceptedBy: this._signerService.did,
+        expirationTimestamp: expiry,
+      });
```
Harasz, line 469: it dont get it, it will always publish role on-chain

jrhender:
@JGiter Could you explain in the PR Description what this PR is intended to accomplish please? I know we discussed possible solutions in the Jira issue, but I'm not sure which one this PR is accomplishing 😅

JGiter:
> @JGiter Could you explain in the PR Description what this PR is intended to accomplish please? I know we discussed possible solutions in the Jira issue, but I'm not sure which one this PR is accomplishing sweat_smile

I updated description

src/modules/claims/claims.service.ts
JGiter: Check is added in registerOnchain method

src/modules/claims/claims.service.ts
```diff
@@ -445,29 +443,30 @@ export class ClaimsService {
       expirationTimestamp: claimExpirationTimestamp,
     };
 
-    if (registrationTypes.includes(RegistrationTypes.OnChain)) {
+    if (
+      registrationTypes.includes(RegistrationTypes.OnChain) &&
+      publishOnChain
```
jrhender, line 448: This might be incorrect because we might want to issue an `onchainProof` but not actually publish (this is what we do for staking KYC and snapshot credentials for example)

jrhender (changes requested):
I'm concerned this might not allow for issuing `onchainProof` without publishing on chain. See comment

src/modules/claims/claims.service.ts
JGiter: Ah, right, I missed that

src/modules/claims/claims.service.ts
```diff
@@ -546,10 +530,23 @@ export class ClaimsService {
...
       claimType,
       acceptedBy,
-      subjectAgreement,
       onChainProof,
       expirationTimestamp,
     } = claim;
+    let { subjectAgreement } = claim;
+
+    if (await this.hasOnChainRole(subject, claimType, +claimTypeVersion)) {
+      return;
```
jrhender, line 539: I think it would be good to add a log line here, maybe at the `info` level.

jrhender (approved):
Looks good, just a suggestion to add a log line

src/modules/claims/claims.service.ts
jrhender: Should add a task to add a test for this case?

JGiter: I would better test this in this PR

src/modules/claims/claims.service.ts
JGiter: I think warning is more appropriate, because it is unexpected

PR #614, opened 2022-07-21
fix: cache server authentication

Harasz:
### Contributor checklist

### DIFF
--- a/src/modules/cacheClient/cacheClient.service.ts
+++ b/src/modules/cacheClient/cacheClient.service.ts
@@ -123,9 +123,14 @@ export class CacheClient implements ICacheClient {
         });
       });
       if (!this.isAuthenticating) {
-        this.isAuthenticating = true;
-        await this.authenticate();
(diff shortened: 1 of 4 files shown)


src/modules/cacheClient/cacheClient.service.ts
```diff
@@ -123,9 +123,14 @@ export class CacheClient implements ICacheClient {
...
       });
       if (!this.isAuthenticating) {
-        this.isAuthenticating = true;
-        await this.authenticate();
-        this.isAuthenticating = false;
+        try {
+          this.isAuthenticating = true;
+          await this.authenticate();
+        } catch {
+          return Promise.reject(error);
```
jrhender, line 130: Do you think it makes sense to add an additional log here? I think it can't hurt, even if it results in the error being logged multiple times.
```suggestion
        } catch {
          console.log("Error while authenticating");
          console.log(error);
          return Promise.reject(error);
```

Harasz:
published

PR #615, opened 2022-07-21
update ew-credentials

JGiter:
### Contributor checklist

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -7,9 +7,7 @@ import {
   PreconditionType,
   RoleCredentialSubject,
 } from '@energyweb/credential-governance';
-import {
-  VerifiableCredential,
-} from '@ew-did-registry/credentials-interface';
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
 import { ClaimRevocation } from '@energyweb/onchain-claims';
 import { Methods } from '@ew-did-registry/did';
 import { Algorithms } from '@ew-did-registry/jwt';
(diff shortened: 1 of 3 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -1540,22 +1533,35 @@ export class ClaimsService {
...
+      this._didRegistry.registrySettings,
+      verifyProof
+    );
+    this._issuerVerification = new IssuerVerification(
+      this._issuerResolver,
+      this._credentialResolver,
+      this._signerService.provider,
+      this._didRegistry.registrySettings,
+      revocationVerification,
+      verifyProof
```
jrhender, line 1564: Do we need to `bind(this)` here so that `this._verifiableCredentialService.verify` in the `verifyProof` implementation works as expected?
```suggestion
      verifyProof.bind(this)
```

JGiter: Hmm, I am not sure why. `this` is already set based on the scope `verifyProof` is defined with, so it will be `ClaimsService`

jrhender: My thinking is that `this` changes in the context of `IssuerVerification`. But actually, I think you're right, because `verifyProof` is an arrow function, I think it captures the value of `this` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions 

> However, because it is a [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures), not the function's own binding, the value of this will not change based on the execution context.

jrhender (approved):
Looks good!

PR #617, opened 2022-07-22
feat: adding more details to error log line when http request fails

artursudnik:
This PR adds more details to `[CACHE CLIENT] axios error:` type error log lines. This includes:
* url
* method
* response status code
* response status text
* response headers
* request headers with `Authorization` header value masked

### DIFF
--- a/src/modules/cache-client/cache-client.service.ts
+++ b/src/modules/cache-client/cache-client.service.ts
@@ -602,6 +602,8 @@ export class CacheClient implements ICacheClient {
         ...error.config.headers,
         Authorization: error.config.headers?.Authorization ? '***' : undefined,
       },
+      status: error.response?.status,
+      statusText: error.response?.statusText,
       responseHeaders: error.response?.headers,
     };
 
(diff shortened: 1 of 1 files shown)


jrhender (approved):
Looks good to me!

src/modules/cache-client/cache-client.service.ts
```diff
@@ -594,7 +594,22 @@ export class CacheClient implements ICacheClient {
       return false;
     }
 
-    getLogger().debug(`[CACHE CLIENT] axios error: ${error.message}`);
+    const errorDetails = {
+      message: error.message,
```
JGiter, line 598: This object can be loged by `inspect` function from utils

JGiter (commented):
May be we should log details only on actual errors

Harasz (approved):

src/modules/cache-client/cache-client.service.ts
artursudnik: I think we need to be kind of backward compatible here because this change is implemented at the request of people struggling while debugging now. So, just adding details to this exact error log line is a good starting point. If it gets too noisy we can narrow it in the next iteration.

PR #618, opened 2022-07-22
fix: set check ethr signer for node clients

Harasz:
### Description

Add missing set for ethr signer for node clients

### Contributor checklist

### DIFF
--- a/src/modules/signer/signer.service.ts
+++ b/src/modules/signer/signer.service.ts
@@ -85,7 +85,10 @@ export class SignerService {
       } else if (isEthSigner === 'false') {
         this._isEthSigner = false;
       }
+    } else {
+      this._setIsEthrSigner();
     }
+
     /**
      * @todo provide general way to initialize with previously saved key
      */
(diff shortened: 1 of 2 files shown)


src/modules/signer/signer.service.ts
```diff
@@ -502,4 +505,43 @@ export class SignerService {
       sig
     )}`;
   }
+
+  /**
+   * Set `_isEthSigner` value based on a signed message.
+   * Generates a test message and signs it.
+   */
+  private async _setIsEthrSigner() {
```
JGiter, line 513: Why not use _calculatePubKeyAndIdentityToken?

Harasz: We can, but _calculatePubKeyAndIdentityToken is making a call to RPC asking about block number, so for performance reason i decide to chose this approach

JGiter (approved):

src/modules/signer/signer.service.ts
```diff
@@ -502,4 +505,43 @@ export class SignerService {
...
+    const encodedHeader = base64url(JSON.stringify(header));
+    const address = this._address;
+    const payload = {
+      address,
+    };
+
+    const encodedPayload = base64url(JSON.stringify(payload));
+    const token = `0x${Buffer.from(
+      `${encodedHeader}.${encodedPayload}`
+    ).toString('hex')}`;
```
jrhender, line 527: I don't think we this JWT stuff. I think we just need to sign a message and see if we can recover the address. Is there an advantage to the JWT bit?

jrhender (changes requested):

src/modules/signer/signer.service.ts
Harasz: Yeah, you are right. simple message is enough

jrhender:
The code looks good, but how hard would it be to add tests for this case?
Sorry, I should have asked this in my initial feedback.

JGiter:
> The code looks good, but how hard would it be to add tests for this case?

I think the simplest way to test is to remove code which currently sets isEthrSigner, like here https://github.com/energywebfoundation/iam-client-lib/blob/e61ff2b7363d6b81b0cb8d34d4d60004a5994c2a/e2e/claims.service.e2e.ts#L208

Harasz:
@jrhender tests added

jrhender (approved):
Looks good, thanks @Harasz

PR #619, opened 2022-07-25
feat(verifyExp): add exp verification method and tests for EIP and VP

whitneypurdum:
### Description

- add method to check if given date is expired (used in verifyRoleEIP191JWT and will be used in Switchboard)
- catch 'Credential Expired' error from verifyVC and push to errors array in verifyVc method
- verify that EIP191JWT is not expired in verifyRoleEIP191JWT. Handle expiration error if so
- Add unit tests for expiration resolveCredentialAndVerify, verifyVc and verifyRoleEIP191JWT

### Contributor checklist

### DIFF
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -948,7 +948,7 @@ describe('Сlaim tests', () => {
             ],
             publishOnChain: true,
             issuerFields: [],
-            expirationTimestamp: Date.now() + 10000,
+            expirationTimestamp: Date.now() + 7000,
           });
           await signerService.connect(rootOwner, ProviderType.PrivateKey);
           const subjectDoc = await didRegistry.getDidDocument({
@@ -960,7 +960,7 @@ describe('Сlaim tests', () => {
             claim: { token: issuedToken },
           });
           const delay = (ms) => new Promise((res) => setTimeout(res, ms));
-          await delay(9000);
+          await delay(8000);
           await signerService.connect(staticIssuer, ProviderType.PrivateKey);
           const result = await claimsService.resolveCredentialAndVerify(
             rootOwnerDID,
@@ -990,13 +990,13 @@ describe('Сlaim tests', () => {
         const vc = await createExampleSignedCredential(
           issuerFields,
           `${vcExpired}.${root}`,
-          new Date(Date.now() + 10000)
+          new Date(Date.now() + 8000)
         );
         nock(vc.credentialStatus?.statusListCredential as string)
           .get('')
           .reply(200, undefined);
         const delay = (ms) => new Promise((res) => setTimeout(res, ms));
-        await delay(11000);
+        await delay(8000);
         const result = await claimsService.verifyVc(vc);
         expect(result.errors).toContain('Verifiable Credential is expired.');
         expect(result.isVerified).toBe(false);
@@ -1012,7 +1012,7 @@ describe('Сlaim tests', () => {
           ],
           publishOnChain: true,
           issuerFields: [],
-          expirationTimestamp: Date.now() + 10000,
+          expirationTimestamp: Date.now() + 7000,
         });
         await signerService.connect(rootOwner, ProviderType.PrivateKey);
         const subjectDoc = await didRegistry.getDidDocument({
@@ -1024,7 +1024,7 @@ describe('Сlaim tests', () => {
           claim: { token: issuedToken },
         });
         const delay = (ms) => new Promise((res) => setTimeout(res, ms));
-        await delay(11000);
+        await delay(8000);
         await signerService.connect(staticIssuer, ProviderType.PrivateKey);
         const credential = await claimsService.fetchCredential(
           rootOwnerDID,
--- a/e2e/claims.service.e2e.ts
+++ b/e2e/claims.service.e2e.ts
@@ -62,8 +62,8 @@ const verifyVcRole = 'verifyVcRole';
 const verifyVcRole2 = 'verifyVcRole2';
 const verifyOffChainClaimRole = 'verifyOnChain';
 const resolveVC = 'resolvevc';
-const verifyVcExpired = 'vcExpired';
-const eipExpired = 'eipExpired';
+const verifyResolvedVcExpired = 'vcResolvedExpired';
+const eip191JwtExpired = 'eip191JwtExpired';
 const vcExpired = 'vcExpired';
 const namespace = root;
 const version = 1;
@@ -119,9 +119,9 @@ const roles: Record<string, IRoleDefinitionV2> = {
     roleName: verifyOffChainClaimRole,
     issuer: { issuerType: 'DID', did: [staticIssuerDID] },
   },
-  [`${verifyVcExpired}.${root}`]: {
+  [`${verifyResolvedVcExpired}.${root}`]: {
     ...baseRoleDef,
-    roleName: verifyVcExpired,
+    roleName: verifyResolvedVcExpired,
     issuer: { issuerType: 'DID', did: [staticIssuerDID] },
   },
   [`${vcExpired}.${root}`]: {
@@ -129,9 +129,9 @@ const roles: Record<string, IRoleDefinitionV2> = {
     roleName: vcExpired,
     issuer: { issuerType: 'DID', did: [staticIssuerDID] },
   },
-  [`${eipExpired}.${root}`]: {
+  [`${eip191JwtExpired}.${root}`]: {
     ...baseRoleDef,
-    roleName: eipExpired,
+    roleName: eip191JwtExpired,
     issuer: { issuerType: 'DID', did: [staticIssuerDID] },
   },
 };
@@ -275,9 +275,9 @@ describe('Сlaim tests', () => {
       returnSteps: false,
     });
     await domainsService.createRole({
-      roleName: verifyVcExpired,
+      roleName: verifyResolvedVcExpired,
       namespace,
-      data: roles[`${verifyVcExpired}.${root}`],
+      data: roles[`${verifyResolvedVcExpired}.${root}`],
       returnSteps: false,
     });
     await domainsService.createRole({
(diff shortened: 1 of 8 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -1494,12 +1505,32 @@ export class ClaimsService {
...
+      }
+    }
     return {
       errors: errors,
-      isVerified: !!proofVerified && !!issuerVerified,
+      isVerified: !!proofVerified && !!issuerVerified && !isExpired,
     };
   }
 
+  async fetchCredential(
```
whitneypurdum, line 1521: I created this method just so that I could grab the credential after I do 'requestAndEnrol', so that I have the cred object to pass to the verifyRoleEIP191JWT. I'm not sure how else I can get the object after its issued otherwise?

src/errors/error-messages.ts
```diff
@@ -35,4 +35,5 @@ export enum ERROR_MESSAGES {
   PROOF_NOT_VERIFIED = 'Proof not verified',
   OFFCHAIN_ISSUER_NOT_AUTHORIZED = 'Issuer of OffChain Claim is not authorized',
   NO_CLAIM_RESOLVED = 'No claim found for given DID and role',
+  CLAIM_EXPIRED = 'Credential Expired',
```
jrhender, line 38:
```suggestion
  CREDENTIAL_EXPIRED = 'Credential Expired',
```

src/modules/claims/claims.service.ts
```diff
@@ -1494,12 +1505,32 @@ export class ClaimsService {
...
+    subjectDID: string,
+    roleNamespace: string
+  ): Promise<
+    VerifiableCredential<RoleCredentialSubject> | RoleEIP191JWT | undefined
+  > {
+    const resolvedCredential = await this._credentialResolver.getCredential(
+      subjectDID,
+      roleNamespace
+    );
+    return resolvedCredential;
```
jrhender, line 1531: It's a bit of personal preference, but it doesn't do much to have the variable in this case. If you like having it, feel free to keep it though.
```suggestion
    return await this._credentialResolver.getCredential(
      subjectDID,
      roleNamespace
    );
```

src/modules/claims/claims.service.ts
```diff
@@ -1438,11 +1441,17 @@ export class ClaimsService {
...
     const issuerDID = this._signerService.did;
 
-    const proofVerified = await this._verifiableCredentialService.verify(vc);
+    let proofVerified;
+    try {
+      proofVerified = await this._verifiableCredentialService.verify(vc);
+    } catch (e) {
+      proofVerified = false;
+      errors.push((e as Error).message);
+    }
```
jrhender, line 1450: This change is a bit odd. Did the behaviour of `verifiableCredentialService.verify` change? Or you just noticed a bug?

src/modules/claims/claims.service.ts
```diff
@@ -1422,6 +1422,9 @@ export class ClaimsService {
       await this._signerService.signMessage(arrayify(proofHash))
     );
   }
+  claimIsExpired(date: number): boolean {
+    return !!date && date < Date.now();
+  }
```
jrhender, line 1427: I think this function is more useful if it checks a JWT payload, rather than just the date check.
Extracting `Date.now()` give more flexibility to the client and make this function pure.

Another question to think about is if `date` should be a `Date` or a `number`

```suggestion
  jwtIsExpired(payload: {Whatever the correct type is for JWT payload}, date: number): boolean {
    return payload?.exp && date && payload.exp < date;
  }
```

e2e/claims.service.e2e.ts
```diff
@@ -61,6 +62,9 @@ const verifyVcRole = 'verifyVcRole';
 const verifyVcRole2 = 'verifyVcRole2';
 const verifyOffChainClaimRole = 'verifyOnChain';
 const resolveVC = 'resolvevc';
+const verifyVcExpired = 'vcExpired';
```
jrhender, line 65: Is this needed if we have `vcExpired` below?
```suggestion
```

jrhender (commented):
Good start but added a few comments. 

IMO, as this PR is missing documentation on some of the functions, it is isn't "Open" (i.e. potentially mergeable) and should be marked as "Draft". I think a PR should only be marked as "Open" if the all the items in the checklist are fullfilled.

src/modules/claims/claims.service.ts
whitneypurdum: The error for the credential being expired is caught and handled here. If I don't catch it, the whole call with throw an exception.

src/modules/claims/claims.service.ts
whitneypurdum: so we never will want to use the date from the VC? Only from the JWT payload?

whitneypurdum: Please let me know if you would rather have a date or a number

jrhender: > so we never will want to use the date from a VC? Only from the JWT payload?

No, we might want to check the expiry of a VC as well. My suggestion would be to have 2 methods:
- jwtIsExpired(jwtPayload: jwt.JwtPayload, dateToCheck: Date)
- vcIsExpired(vc: VerifiableCredential, dateToCheck: Date)

> Please let me know if you would rather have a date or a number

I think it is better to use a `Date` type as then there is no ambiguity on what the number should be (e.g. seconds/miliseconds, since when, etc)

src/modules/claims/claims.service.ts
jrhender: Ah okay, makes sense, thanks.

src/modules/claims/claims.service.ts
jrhender: By the way, I was reading this SO answer that is saying that `exp` from a JWT needs to be multiplied for the check https://stackoverflow.com/a/51293316/6674421

jrhender (commented):
A few further comments but nearly there I think. Thanks @whitneypurdum !

src/modules/claims/claims.service.ts
whitneypurdum: But in the use case of switchboard, we won't be passing either a JWT OR a VC to this method - we will just be pulling the date from expirationTimestamp from the claim as it comes to switchboard
`expirationTimestamp: "1658847564357"`

- so can we have another method that just takes in a date number and does a check? And then a method for jwtIsExpired etc...

jrhender: @whitneypurdum 
> we will just be pulling the date from expirationTimestamp from the claim as it comes to switchboard
expirationTimestamp: "1658847564357"

Fair enough, I didn't realize this part. I think we should delete this function then from ICL. The best code is no code at all 😄

e2e/claims.service.e2e.ts
```diff
@@ -913,6 +984,58 @@ describe('Сlaim tests', () => {
...
+        const vc = await createExampleSignedCredential(
+          issuerFields,
+          `${vcExpired}.${root}`,
+          new Date(Date.now() + 10000)
+        );
+        nock(vc.credentialStatus?.statusListCredential as string)
+          .get('')
+          .reply(200, undefined);
+        const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+        await delay(11000);
```
jrhender, line 999: A minor point, but does the delay need to be this long? 11sec is quite a bit of time for a test. Would 1 or 2 seconds (or even less) do?

e2e/claims.service.e2e.ts
```diff
@@ -913,6 +984,58 @@ describe('Сlaim tests', () => {
...
+        const subjectDoc = await didRegistry.getDidDocument({
+          did: rootOwnerDID,
+          includeClaims: true,
+        });
+        mockCachedDocument.mockResolvedValueOnce(subjectDoc);
+        await claimsService.publishPublicClaim({
+          claim: { token: issuedToken },
+        });
+        const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+        await delay(11000);
```
jrhender, line 1027: Can we also reduce this delay?

src/modules/claims/claims.service.ts
whitneypurdum: @jrhender so in the verifyRoleEIP191JWT method, just verify the expiration in the method itself (just a basic number comparison?)

e2e/claims.service.e2e.ts
whitneypurdum: I'll try decreasing the numbers. The expiration date is set to be 11 seconds from the time of issuance to allow time for issuance etc,  so I just want to make sure that we are doing the check after that time passes. Will try to decrease.

src/modules/claims/claims.service.ts
jrhender: Yes, that's what I'm thinking for now. If we see room for refactor, maybe we do that down the line

whitneypurdum:
Looks like you can actually verify if a token is expired by calling jwt.verify(token) https://github.com/auth0/node-jsonwebtoken/blob/master/README.md#jwtverifytoken-secretorpublickey-options-callback

src/modules/claims/claims.service.ts
```diff
@@ -1494,12 +1501,35 @@ export class ClaimsService {
     if (!proofVerified) {
       errors.push(ERROR_MESSAGES.PROOF_NOT_VERIFIED);
     }
+    const isExpired = payload?.exp && payload?.exp * 1000 < Date.now();
```
jrhender, line 1504: I think this is a nice explainer to have here
```suggestion
    // Date.now() and JWT expiration time both identify the time elapsed since January 1, 1970 00:00:00 UTC
    const isExpired = payload?.exp && payload.exp * 1000 < Date.now();
```

jrhender (approved):
Looks good to me, thanks @whitneypurdum !

src/modules/claims/claims.service.ts
```diff
@@ -1494,12 +1501,36 @@ export class ClaimsService {
...
     }
+    // Date.now() and JWT expiration time both identify the time elapsed since January 1, 1970 00:00:00 UTC
+    const isExpired = payload?.exp && payload?.exp * 1000 < Date.now();
+    if (isExpired) {
+      errors.push(ERROR_MESSAGES.CREDENTIAL_EXPIRED);
+    }
     return {
       errors: errors,
-      isVerified: !!proofVerified && !!issuerVerified,
+      isVerified: !!proofVerified && !!issuerVerified && !isExpired,
```
JGiter, line 1511: `issuerVerified` seem to be boolean
```suggestion
      isVerified: !!proofVerified && issuerVerified && !isExpired,
```

JGiter (approved):

PR #621, opened 2022-07-27
task/ICL-126 session management

JGiter:
### Contributor checklist

- [X] Breaking changes - check for any existing interfaces changes that are not backward compatible, removed method etc.
- [X] Tests - add new or updated existed test for changes you made.
- [X] Migration guide - add migration guide for every breaking change.
- [X] Configuration correctness - check that any configuration changes are correct ex. default URLs, chain ids, smart contract verification on [Volta explorer](https://volta-explorer.energyweb.org/) or [EWC explorer](https://explorer.energyweb.org/).

### DIFF
--- a/docs/setup/using_in_browser.md
+++ b/docs/setup/using_in_browser.md
@@ -1,0 +1,11 @@
+[Initialization of library](./intialization.md) requires several properties of [signer](../api/classes/modules_signer.SignerService.md):
+
+- The signer meets [EIP-191: Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191)
+- Public key. This is legacy parameter, which was used to initialized DID document
+
+These properties are not directly exposed by all signers, such as Metamask, and thus are determined in runtime by sending test sign request.
+In order to improve user expirience when library is used in browser it tries to get these parameters from local storage.
+In order to implement application specific session management the library doesn't modify storage and relies on the application to set `isEthSigner` and `PublicKey` local storage variables on login.
+For example to disable session application will not set properties.
+If properties are not in local storage then signer initialization will be completed on login to ssi-hub.
+Initialization is postponed in order to prevent double signing when user is not logged in ssi-hub.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -6,6 +6,7 @@ nav:
   - Setup:
       - Initialization: setup/initialization.md
       - Logging: setup/logging.md
+      - Using in browser: setup/using_in_browser.md
   - Guides:
       - Application: guides/application.md
       - Asset: guides/asset.md


docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,6 @@
+Initialization of library requires several properties of signer:
+
+- Is signer meets ethereum signer specification https://eips.ethereum.org/EIPS/eip-191
```
jrhender, line 3: I think this is a better sentence formulation
```suggestion
- The signer meets [EIP-191: Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191)
```

docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,6 @@
...
+- Public key. This is legacy parameter, which was used to initialized DID document
+
+These properties are not directly exposed by all signers and thus are determined in runtime by sending test sign request. In order to improve user expirience when library is used in browser it tries to get these parameters from local storage. Because library doesn't change storage application should set `isEthSigner` and `PublicKey` storage variables on login. Library doesn't send sign request if properties are not in local storage. Othewise user will have to sign twice if he was not logged in to cache server
```
jrhender, line 6: Does you have an example signer for which this isn't met?

Also, I also think markdown is a bit easier to maintain if each sentence is split into a separate line. This doesn't affect the rendered results but makes it easier to move between sentences and move them around.
```suggestion
These properties are not directly exposed by all signers and thus are determined in runtime by sending test sign request.
In order to improve user expirience when library is used in browser it tries to get these parameters from local storage.
Because library doesn't change storage application should set `isEthSigner` and `PublicKey` storage variables on login.
Library doesn't send sign request if properties are not in local storage. 
Otherwise user will have to sign twice if he was not logged in to cache server
```

jrhender (commented):
Some comments

docs/setup/using_in_browser.md
JGiter: > Does you have an example signer for which this isn't met?

I'll give Metamask as an example

docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,10 @@
+[Initialization of library](./intialization.md) requires several properties of [signer](../api/classes/modules_signer.SignerService.md):
+
+- The signer meets [EIP-191: Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191)
+- Public key. This is legacy parameter, which was used to initialized DID document
+
+These properties are not directly exposed by all signers, such as Metamask, and thus are determined in runtime by sending test sign request.
+In order to improve user expirience when library is used in browser it tries to get these parameters from local storage.
+Because library doesn't change storage application should set `isEthSigner` and `PublicKey` storage variables on login.
```
jrhender, line 8: What do you mean by "the library doesn't change storage" here? I guess you mean that application javascript can't set local storage? I don't think this is the case (https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem), so I'm a bit confused as to what this line means.
```suggestion
Because the library doesn't change storage, the application should set `isEthSigner` and `PublicKey` local storage variables on login.
```

docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,10 @@
...
+Library doesn't send sign request if properties are not in local storage.
```
jrhender, line 9: Shouldn't it be the other way around. The library **does** send the request if properties are not in local storage?

docs/setup/using_in_browser.md
JGiter: Yes, I meant local storage. By 
> the library doesn't change storage
I mean we prefer that the library doesn't modify the local storage and let the application manage the session as needed. I will paraphrase this

docs/setup/using_in_browser.md
JGiter: Yes, it is send request, but not as part of signer initialization. `isEthrSigner` and `publicKey` are calculated as part of login to cache server. I'll make this more clear

JGiter: Maybe we want to change that and initialize signer completely before login to ssi-hub

docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,10 @@
+[Initialization of library](./intialization.md) requires several properties of [signer](../api/classes/modules_signer.SignerService.md):
+
+- The signer meets [EIP-191: Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191)
+- Public key. This is legacy parameter, which was used to initialized DID document
+
+These properties are not directly exposed by all signers, such as Metamask, and thus are determined in runtime by sending test sign request.
+In order to improve user expirience when library is used in browser it tries to get these parameters from local storage.
+In order to implement application specific session management the library doesn't modify storage and rely on the application to set `isEthSigner` and `PublicKey` local storage variables on login.
+If properties are not in local storage then signer initialization will be completed on login to ssi-hub.
+Initialization is posponed in order to prevent double signing when user is not logged in ssi-hub.
```
jrhender, line 10:
```suggestion
Initialization is postponed in order to prevent double signing when user is not logged in ssi-hub.
```

docs/setup/using_in_browser.md
```diff
@@ -0,0 +1,10 @@
+[Initialization of library](./intialization.md) requires several properties of [signer](../api/classes/modules_signer.SignerService.md):
+
+- The signer meets [EIP-191: Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191)
+- Public key. This is legacy parameter, which was used to initialized DID document
+
+These properties are not directly exposed by all signers, such as Metamask, and thus are determined in runtime by sending test sign request.
+In order to improve user expirience when library is used in browser it tries to get these parameters from local storage.
+In order to implement application specific session management the library doesn't modify storage and rely on the application to set `isEthSigner` and `PublicKey` local storage variables on login.
```
jrhender, line 8:
```suggestion
In order to implement application specific session management the library doesn't modify storage and relies on the application to set `isEthSigner` and `PublicKey` local storage variables on login.
```

jrhender: I don't rely understand this logic. We want "application specific session management" so "the library doesn't modify storage", but the library is relying on "local storage variables".
So we don't allow the library to write to local storage but the library relies on local storage, so it seems the use of local storage for session management is forced by the library. Maybe you could add an example of how "application specific session management" would be used?

JGiter: Library doesn't force use of local storage. Application can refuse from session at all and library will calculate signer properties on each login

jrhender (approved):
A couple typos to fix still but I think this is helpful documentation to have, so let's get it in!

PR #623, opened 2022-08-03
fix(exp): do not use dflt validity period to calculate timestamp if n…

whitneypurdum:
### Description

- Remove logic to calculate timestamp from default validity period if no expiration timestamp is sent AND role has a default validity period

### Contributor checklist


jrhender (approved):
Looks good, thanks @whitneypurdum

PR #627, opened 2022-08-05
bug/ICL-346 can't add public key to asset document

JGiter:
### Contributor checklist


JGiter:
The fix isn't completed with test, because we don't have test setup for did-registry :disappointed:

jrhender (approved):

PR #630, opened 2022-08-09
fix: use resolveCredentialAndVerify to verify enrolment prerequisites

whitneypurdum:
### Description
Addresses bug https://energyweb.atlassian.net/jira/software/projects/ICL/boards/85?selectedIssue=ICL-345
- Instead of using a user's claims from getUserClaims (which only returns claims where the currentUser is the issuer), this verifies that a user has the necessary role pre-requisites using resolveCredentialAndVerify. 
- Also updates resolveCredentialAndVerify to use the issuer did of the credential, instead of using the current signer DID

### Contributor checklist


jrhender (commented):
@whitneypurdum do you know why the documentation changes here? They seem unrelated to the substantive changes that are being made

src/modules/claims/claims.service.ts
```diff
@@ -1448,7 +1448,7 @@ export class ClaimsService {
     const role = vc.credentialSubject.role.namespace;
     let issuerVerified = true;
     try {
-      await this._issuerVerification.verifyIssuer(issuerDID, role);
+      await this._issuerVerification.verifyIssuer(issuerDID as string, role);
```
jrhender, line 1451: Rather than use `as string`, I think it's better to add a check after `const issuerDID = vc.issuer` so that error can be thrown closer to the source.

src/modules/claims/claims.service.ts
```diff
@@ -1481,10 +1481,10 @@ export class ClaimsService {
   ): Promise<CredentialVerificationResult> {
     const { payload, eip191Jwt } = roleEIP191JWT;
     const errors: string[] = [];
-    const issuerDID = this._signerService.did;
+    const issuerDID = roleEIP191JWT.payload.iss;
     const { status: issuerVerified, error } =
       await this._issuerVerification.verifyIssuer(
-        issuerDID,
+        issuerDID as string,
```
jrhender, line 1487: Similar to above, I think it's better to throw an error than pass a possibly undefined value into a function that expects a string
```suggestion
        issuerDID,
```

whitneypurdum:
> @whitneypurdum do you know why the documentation changes here? They seem unrelated to the substantive changes that are being made

Yes, my bad! I made some changes locally when I thought we would use the gerServices endpoint, and I think the build changed the documentation. I will remove.

jrhender (commented):
Some comments but it looks like a good direction to me

Are you able to test and confirm that this allows an issuer to issue without being an authorized issuer of the prerequisite role?

src/modules/claims/claims.service.ts
whitneypurdum: @jrhender Should we throw an error if no issuer, or just add to the array of errors that there is no issuer and carry on with verification (and not call verify issuer)

jrhender: I guess I would add to the error array and return. If there is no issuer then doing any verification doesn't make sense as it is a required property of the VC data model https://www.w3.org/TR/vc-data-model/#issuer

whitneypurdum: Oh okay - then it that case maybe it would make sense to throw so we can stop the process?

jrhender: Right, if that makes most sense to you, I could see that working. Both ways make sense to me

whitneypurdum:
> Some comments but it looks like a good direction to me
> 
> Are you able to test and confirm that this allows an issuer to issue without being an authorized issuer of the prerequisite role?

I added a test specifically for this.

src/modules/claims/claims.service.ts
whitneypurdum: Throwing an error now. vc.issuer could be string or IIssuer, but when I check for vc.issuer.id, I get a type error? Only vc.issuer as string is valid. 
export interface IIssuer {
    id: string;
    [x: string]: unknown;
}

src/modules/claims/claims.service.ts
```diff
@@ -1250,21 +1250,21 @@ export class ClaimsService {
     const { enrolmentPreconditions } = roleDefinition as IRoleDefinition;
```
nichonien, line 1250:
```suggestion
    const { enrolmentPreconditions } = roleDefinition as IRoleDefinitionV2;
```
@whitneypurdum The IRoleDefinitionV2 is the interface with `revokers` and `defaultValidityPeriod` https://github.com/energywebfoundation/ew-credentials/blob/77bd399bbbe78e32f85a4884e4e7073bdeff0dee/packages/credential-governance/src/types/domain-definitions.ts#L23

nichonien: I think we should do a type check, as the roleDefinition coming from `this._domainsService.getDefinition` returns `Promise< IRoleDefinition | IRoleDefinitionV2 | IAppDefinition | IOrganizationDefinition`. https://github.com/energywebfoundation/iam-client-lib/blob/ff6886c517bc55195691d1d316b770afa4dffde7/src/modules/domains/domains.service.ts#L802

nichonien: @jrhender @JGiter We should create separate methods to get definitions, fetching everything with one function `domainsService.getDefinition` can be confusing and would need type validation every time. What are your thoughts?

JGiter (commented):
Does this fix mean that credential is in document, but is missing in ssi-hub?

src/modules/claims/claims.service.ts
JGiter: Make sense to me

jrhender:
> Does this fix mean that credential is in document, but is missing in ssi-hub?

Possibly (in the future we may want to use different ssi-hub's for different credential exchanges)

The more likely case is the the issuer of the current credential just isn't authorized to retrieve the prerequisite credential from ssi-hub

JGiter:
> The more likely case is the the issuer of the current credential just isn't authorized to retrieve the prerequisite credential from ssi-hub

So can we think that this is bug in ssi-hub, which prevents legal issuer to verify prerequisites?

src/modules/claims/claims.service.ts
whitneypurdum: Thanks @nichonien I updated that, I think I just need enrolmentPreconditions, which is on both interfaces though.

src/modules/claims/claims.service.ts
jrhender: Can the cast be removed with the above check? I would think that it could be
```suggestion
      await this._issuerVerification.verifyIssuer(issuerDID, role);
```

jrhender: How about this change? I think it is better than the cast as before the code would fail if the type was indeed `IIssuer`
```suggestion
      if (typeof issuerDID === 'string') {
        await this._issuerVerification.verifyIssuer(issuerDID, role);
      }
      else {
        await this._issuerVerification.verifyIssuer(issuerDID.id, role);
      }
```

src/modules/claims/claims.service.ts
```diff
@@ -1252,24 +1252,24 @@ export class ClaimsService {
       throw new Error(ERROR_MESSAGES.ROLE_NOT_EXISTS);
     }
 
-    const { enrolmentPreconditions } = roleDefinition as IRoleDefinition;
+    const { enrolmentPreconditions } = roleDefinition as IRoleDefinitionV2;
```
nichonien, line 1255: @whitneypurdum can we put a check here something like :

if (DomainReader.isRoleDefinitionV2(roleDefinition)) {
      some code;
    }

jrhender: @nichonien what should the behaviour be if the `roleDefinition` is not `IRoleDefinitionV2`? Should an error be thrown? I don't think this is necessary because both `IRoleDefinition` and `IRoleDefinitionV2` have `enrolmentPreconditions`, right? We could technically have the code below I think 
```suggestion
    const { enrolmentPreconditions } = roleDefinition as IRoleDefinition | IRoleDefinitionV2;
```

jrhender (approved):

src/modules/claims/claims.service.ts
nichonien: @jrhender The roleDefinition coming from `this._domainsService.getDefinition` returns `Promise< IRoleDefinition | IRoleDefinitionV2 | IAppDefinition | IOrganizationDefinition`. So I think it would be better to do a type check.

jrhender: @nichonien Good point. I think it should be done as a separate PR though as this `as RoleDefinition` cast was already happening prior to this PR. How does that sound?

nichonien: Yes, sounds good to me. Also, we should break `this. _domainsService.getDefinition` into modular methods to get specific definitions.

nichonien (approved):

PR #631, opened 2022-08-09
fix(exp): document expirationTimestamp

whitneypurdum:
### Description

Updates documentation for expirationTimestamp to be explicit that the timestamp is in milliseconds

### Contributor checklist

### DIFF
--- a/docs/api/interfaces/modules_claims.IssueClaimOptions.md
+++ b/docs/api/interfaces/modules_claims.IssueClaimOptions.md
@@ -40,7 +40,7 @@ ___
 
 • `Optional` **expirationTimestamp**: `number`
 
-Defines how long the claim is valid (milliseconds)
+A Unix timestamp expressed in milliseconds of when the claim expires.
 
 ___
 
--- a/docs/api/interfaces/modules_claims.IssueClaimRequestOptions.md
+++ b/docs/api/interfaces/modules_claims.IssueClaimRequestOptions.md
@@ -28,7 +28,7 @@ ___
 
 • `Optional` **expirationTimestamp**: `number`
 
-Defines how long the claim is valid.
+A Unix timestamp expressed in milliseconds of when the claim expires.
 
 ___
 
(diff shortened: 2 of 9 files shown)


nichonien (approved):

jrhender (changes requested):
We should specific that it is a Unix timestamp (which I think is the case), the start date from which the milliseconds are added is important

jrhender:
Actually, I see that a Unix timestamp is in seconds not miliseconds, so we can't say that is directly a Unix timestamp. Perhaps we should say that it is a "Unix timestamp expressed in milliseconds"?

jrhender:
@whitneypurdum In this PR, @JGiter is using seconds as the unit for `RegisterOnchainOptions` which I think might be correct https://github.com/energywebfoundation/iam-client-lib/pull/636/files 
What are your thoughts?

whitneypurdum:
@jrhender so just in registerOnChainClaim
Options is it seconds?

whitneypurdum:
> @whitneypurdum In this PR, @JGiter is using seconds as the unit for `RegisterOnchainOptions` which I think might be correct https://github.com/energywebfoundation/iam-client-lib/pull/636/files What are your thoughts?

updated so  RegisterOnChainOptions is in seconds.

jrhender:
@whitneypurdum I'm sorry, I'm looking at it again in and I think maybe it should be milliseconds in `RegisterOnchainOptions` and it is wrong in https://github.com/energywebfoundation/iam-client-lib/pull/636/files 
Maybe we merge with milliseconds in this PR and then we can debate seconds vs milliseconds in the other PR 😄

whitneypurdum:
> @whitneypurdum I'm sorry, I'm looking at it again in and I think maybe it should be milliseconds in `RegisterOnchainOptions` and it is wrong in https://github.com/energywebfoundation/iam-client-lib/pull/636/files Maybe we merge with milliseconds in this PR and then we can debate seconds vs milliseconds in the other PR 😄

OKay @jrhender , I changed it back to milliseconds. We can't merge anyways till we figure out why the tests are failing. This PR doesn't have the updated packages and it is now failing. Maybe something wrong with cache server?

whitneypurdum:
@jrhender Please can you review this one more time?

jrhender (approved):
looks good. Could you fix the merge conflicts when you have a moment please @whitneypurdum ?

PR #634, opened 2022-08-10
bug/incorrect onchain claim expiration

JGiter:
### Contributor checklist


src/modules/claims/claims.service.ts
```diff
@@ -683,11 +685,11 @@ export class ClaimsService {
 
     if (registrationTypes.includes(RegistrationTypes.OnChain)) {
       const { claimType: role, claimTypeVersion: version } = claim;
-      const expiry = expirationTimestamp || eternityTimestamp;
+      const expiry = Math.floor((expirationTimestamp || eternityTimestamp)/1000);
```
jrhender, line 688: I think there is no need to divide the `eternityTimestamp` by 1000. It should still be the maximum value, IMO
```suggestion
      const expiry = Math.floor(expirationTimestamp / 1000) || eternityTimestamp;
```

src/modules/claims/claims.service.ts
```diff
@@ -683,11 +685,11 @@ export class ClaimsService {
...
       const onChainProof = await this.createOnChainProof(
         role,
         version,
-        expiry,
+        Math.floor(expiry / 1000),
```
jrhender, line 692: Isn't this a second division by 1000? Isn't it being done already a couple lines above?

PR #638, opened 2022-08-11
fix(infura): allow DID reg to take in Infura config

whitneypurdum:
### Description

Allow DID Registry to take in Infura config
### Contributor checklist


e2e/cache-client.e2e.ts
```diff
@@ -18,7 +18,22 @@ describe.skip('Cache client', () => {
     );
     const { connectToDidRegistry, assetsService } =
       await connectToCacheServer();
-    const { claimsService, didRegistry } = await connectToDidRegistry();
+    const projectId = '2DCe3TNAHzzmoSBrcPyWx4NOyrv';
```
whitneypurdum, line 21: I will change these configs - what configs should we be using for test files? I will make helper and share among e2e tests.

src/modules/did-registry/did.types.ts
```diff
@@ -12,6 +12,13 @@ export interface AssetProfile {
   icon?: string;
 }
 
+export interface IpfsConfig {
```
whitneypurdum, line 15: I think this is the best file for this interface, but have doubts

whitneypurdum:
I am not running the tests in case it is not good to use the config that's hard coded.

e2e/cache-client.e2e.ts
jrhender: Best to use
ProjectID: 2DD5GcJRpMeuqMoKsgT95AQ8OI3
ProjectSecret: 842995992aa1edd763196b30d727a45d

jrhender:
> I am not running the tests in case it is not good to use the config that's hard coded.

I think ideally, it would be possible to run the tests against a mock IPFS `ipfsDaemon`.
But we can use actually project values for this first PR if that is easier.

src/init.ts
```diff
@@ -98,13 +98,13 @@ export async function init(signerService: SignerService) {
     );
 
     async function connectToDidRegistry(
-      ipfsStore?: string
+      ipfsConfig: IpfsConfig
```
whitneypurdum, line 101: Is there a reason why this should stay optional? Should there be a default config set in ICL in case it is not passed in?

jrhender (approved):

nichonien (approved):

PR #640, opened 2022-08-18
feat: update ewc claimrevocationregistry addr

nichonien:
### Description

Update EWC ClaimRevocationRegistry contract address

### DIFF
--- a/src/config/chain.config.ts
+++ b/src/config/chain.config.ts
@@ -78,7 +78,8 @@ const chainConfig: Record<number, ChainConfig> = {
     claimManagerAddress: EWC_CLAIM_MANAGER_ADDRESS,
     stakingPoolFactoryAddress: '', // This is not deployed on EWC
     credentialRevocationRegistryAddress: '', // This is not deployed on EWC
-    claimsRevocationRegistryAddress: '', // This is not deployed on EWC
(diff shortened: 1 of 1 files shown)


src/config/chain.config.ts
```diff
@@ -78,7 +78,8 @@ const chainConfig: Record<number, ChainConfig> = {
     claimManagerAddress: EWC_CLAIM_MANAGER_ADDRESS,
     stakingPoolFactoryAddress: '', // This is not deployed on EWC
     credentialRevocationRegistryAddress: '', // This is not deployed on EWC
-    claimsRevocationRegistryAddress: '', // This is not deployed on EWC
+    claimsRevocationRegistryAddress:
```
JGiter, line 81: Is it possible to export it from did-reg? I think we will need this in other packages like passport-did-auth

nichonien: I have added this here just to test it initially, it would definitely come from `ew-credentials` - https://github.com/energywebfoundation/ew-credentials/blob/develop/packages/credential-governance/src/chain-constants.ts

JGiter (approved):

nichonien:
This task is done  as part of https://github.com/energywebfoundation/iam-client-lib/pull/649

PR #644, opened 2022-08-30
#swtch1425 - use subject DID to for getClaimsBySubject

whitneypurdum:
### Description

- Use the subject DID to retrieve claimsBySubject (rather than defaulting to signer DID)

### Contributor checklist

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -806,7 +806,9 @@ export class ClaimsService {
       if (!claim.token) {
         throw new Error(ERROR_MESSAGES.CLAIM_DOES_NOT_CONTAIN_TOKEN);
       }
-      this._didRegistry.isClaim(payload);
+      if (!this._didRegistry.isClaim(payload)) {
+        throw new Error(`${ERROR_MESSAGES.CLAIM_TOKEN_DATA_MISSING}`);
+      }
       const token = claim.token as string;
       const verifiedDid = await this._didRegistry.verifyPublicClaim(
         token,
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -757,6 +757,9 @@ export class ClaimsService {
     registrationTypes = [RegistrationTypes.OffChain],
     claim,
   }: PublishPublicClaimOptions): Promise<string | undefined> {
+    if (!claim.token || token) {
+      throw new Error(ERROR_MESSAGES.CLAIM_DOES_NOT_CONTAIN_TOKEN);
+    }
     claim.token = claim.token || (token as string);
     const payload = (await this._didRegistry.decodeJWTToken({
       token: claim.token,
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -757,7 +757,7 @@ export class ClaimsService {
     registrationTypes = [RegistrationTypes.OffChain],
     claim,
   }: PublishPublicClaimOptions): Promise<string | undefined> {
-    if (!claim.token || token) {
+    if (!claim.token && !token) {
       throw new Error(ERROR_MESSAGES.CLAIM_DOES_NOT_CONTAIN_TOKEN);
     }
     claim.token = claim.token || (token as string);
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -757,9 +757,6 @@ export class ClaimsService {
     registrationTypes = [RegistrationTypes.OffChain],
     claim,
   }: PublishPublicClaimOptions): Promise<string | undefined> {
-    if (!claim.token && !token) {
-      throw new Error(ERROR_MESSAGES.CLAIM_DOES_NOT_CONTAIN_TOKEN);
-    }
     claim.token = claim.token || (token as string);
     const payload = (await this._didRegistry.decodeJWTToken({
       token: claim.token,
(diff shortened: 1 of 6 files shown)


whitneypurdum:
@JGiter I've run these tests four times locally and they always pass, but they consistently fail here due to timeouts. Have you ever come across this? Not sure how to fix as can't reproduce the issue locally.

src/modules/claims/claims.service.ts
```diff
@@ -757,7 +757,20 @@ export class ClaimsService {
     registrationTypes = [RegistrationTypes.OffChain],
     claim,
   }: PublishPublicClaimOptions): Promise<string | undefined> {
-    claim.token = claim.token || token;
+    claim.token = claim.token || (token as string);
```
JGiter, line 760: More safer to throw error if neither `claim.token` nor `token` provided

src/modules/claims/claims.service.ts
```diff
@@ -757,7 +757,20 @@ export class ClaimsService {
...
+    const payload = (await this._didRegistry.decodeJWTToken({
```
JGiter, line 761: It make sense to decode and validate token in separate method, because such operation is performed in several places

src/modules/claims/claims.service.ts
```diff
@@ -791,18 +804,6 @@ export class ClaimsService {
     // can we break API so that register on chain required only claim type and claim type version and subject
     if (registrationTypes.includes(RegistrationTypes.OffChain)) {
       const token = claim.token as string;
-      const payload = (await this._didRegistry.decodeJWTToken({ token })) as {
```
JGiter, line 794: Do you think we should check if subject is current DID?

JGiter:
> @JGiter I've run these tests four times locally and they always pass, but they consistently fail here due to timeouts. Have you ever come across this? Not sure how to fix as can't reproduce the issue locally.

Try to increase timeout

src/modules/claims/claims.service.ts
whitneypurdum: Does the behavior change if it is one vs the other? @JGiter

JGiter: It is just a matter of error handling. Currently did-registry doesn't handles failed update of document. Debugging would be easer if iam-lib prevent such error with explanatory message

src/modules/claims/claims.service.ts
JGiter: I would make it part of `DidRegistry.decodeJwtToken`

src/modules/claims/claims.service.ts
whitneypurdum: But I don't think it should be an error if the subject isn't the current DID. In the case of an asset it wouldn't be. Maybe I'm not following what you are saying though.

JGiter: Ah, right. `updateDocument` checks if document is controlled

src/modules/claims/claims.service.ts
whitneypurdum: @JGiter is this what you were thinking?

src/modules/claims/claims.service.ts
whitneypurdum: I'm not sure what you mean about validate token. It is already decoding in that decodeJwtToken helper.

JGiter: By validation I mean checking that payload has `iss` and `claimData`

whitneypurdum: @JGiter and are you saying you want to throw an error if those are not found?

JGiter: Yes, `publishPublicClaim` expect `iss` and `claimData` in `payload`

whitneypurdum: > I would make it part of `DidRegistry.decodeJwtToken`

This method is used for on and off chain claims, and a claim that is on-chain registration only will not have a token to decode. The test for 'should be able to issue and publish onchain without request' would not pass as an error would be thrown that there is no token to validate. I put this in the off-chain logic only, and made a separate validate method for the decoded token.

JGiter (changes requested):

src/modules/claims/claims.service.ts
```diff
@@ -790,28 +803,23 @@ export class ClaimsService {
...
-      let sub = payload.sub;
-      // Initialy subject was ignored because it was requester
-      if (!sub || sub.length === 0 || !isValidDID(sub)) {
-        sub = this._signerService.did;
+      if (!claim.token) {
+        throw new Error(ERROR_MESSAGES.CLAIM_DOES_NOT_CONTAIN_TOKEN);
       }
-      const verifiedDid = await this._didRegistry.verifyPublicClaim(token, iss);
-      if (!verifiedDid || !compareDID(verifiedDid, iss)) {
+      this._didRegistry.isClaim(payload);
```
JGiter, line 809: I think it is supposed to be used like this
```suggestion
      if (!this._didRegistry.isClaim(payload)) {
         throw new Error(...)
       };
```

JGiter (changes requested):

src/modules/claims/claims.service.ts
```diff
@@ -790,28 +803,25 @@ export class ClaimsService {
...
-        sub: string;
-
-        claimData: ClaimData;
-      };
-      const { iss, claimData } = payload;
-      let sub = payload.sub;
-      // Initialy subject was ignored because it was requester
-      if (!sub || sub.length === 0 || !isValidDID(sub)) {
-        sub = this._signerService.did;
+      if (!claim.token) {
```
JGiter, line 806: Does line 760 ensures that `claim.token` is defined?

JGiter (approved):
Thanks, looks fine to me

PR #645, opened 2022-08-31
fix: validate issuer verification

nichonien:
### Description

This PR validates the issuer verification result and throws if the verification failed for some reason.


src/modules/claims/claims.service.ts
```diff
@@ -1227,8 +1235,11 @@ export class ClaimsService {
    *
    * @param {String} role Registration types of the claim
    */
-  private async verifyIssuer(role: string): Promise<void> {
-    await this._issuerVerification.verifyIssuer(this._signerService.did, role);
+  private async verifyIssuer(role: string): Promise<VerificationResult> {
```
JGiter, line 1238: There is not much sense in this method now, because it is just forwards call. How about throwing here instead of in calling code?

src/modules/claims/claims.service.ts
```diff
@@ -1563,9 +1574,9 @@ export class ClaimsService {
         errors: [ERROR_MESSAGES.NO_CLAIM_RESOLVED],
       };
     }
-    const credentialIsOffChain = resolvedCredential.eip191Jwt;
+    const credentialIsOffChain = isEIP191Jwt(resolvedCredential);
```
JGiter, line 1577: Is it possible to do without extra variable and casting, like so:
```
if (isEIPjwt(resolvedCredential) {
   return this.verifyRoleEIP191JWT(resolvedCredential);
} 
else if isVerifiableCredential(resolvedCredential) {
   return this.verifyVc(resolvedCredential);
}
```

JGiter (changes requested):

src/modules/claims/claims.service.ts
nichonien: I think this would increase if conditions for no reasons, I could move this to ternary operation like 
`return isEIP191Jwt(resolvedCredential) ? this.verifyRoleEIP191JWT(resolvedCredential)`

src/modules/claims/claims.service.ts
```diff
@@ -1574,8 +1571,7 @@ export class ClaimsService {
         errors: [ERROR_MESSAGES.NO_CLAIM_RESOLVED],
       };
     }
-    const credentialIsOffChain = isEIP191Jwt(resolvedCredential);
-    return credentialIsOffChain
+    return isEIP191Jwt(resolvedCredential)
       ? this.verifyRoleEIP191JWT(resolvedCredential)
       : this.verifyVc(
```
JGiter, line 1576: I would also verify vc, sense we created guard for this

JGiter (approved):

src/modules/claims/claims.service.ts
nichonien: This is just a validation to make calls, the credentials are type verified during issuer verification.

PR #649, opened 2022-09-05
task/ICL-359 import revocation registry address

JGiter:
### Description

Hardcoded claim revocation registry address replaced by imported

### Contributor checklist

- [X] Breaking changes - check for any existing interfaces changes that are not backward compatible, removed method etc.
- [X] Documentation - document your code, add comments for method, remember to check if auto generated docs were updated.


src/modules/claims/claims.service.ts
```diff
@@ -674,12 +673,12 @@ export class ClaimsService {
       const publicClaim: IPublicClaim = {
         did: subject,
         signer: this._signerService.did,
+        exp: expirationTimestamp,
```
jrhender, line 676: Are the units correct here? `exp` should be in seconds (https://www.npmjs.com/package/jsonwebtoken#token-expiration-exp-claim) but I think `expirationTimestamp` might be in milliseconds?

jrhender: I think that maybe this line in the tests is incorrect or needs to be updated: https://github.com/energywebfoundation/iam-client-lib/blob/da433e54d2b23c36d40e1552a0679aba6af8cb86/e2e/claims.service.e2e.ts#L498

JGiter: Looks like this a bug. I'll move this in another issue

nichonien: @JGiter @jrhender The current implementation is such that the expiry for on chain claims is in seconds while the `roleEIP191Jwt` and `VC` are in milliseconds, this is to align expiration CRUD operation in VC and EIP191Jwt.

nichonien (approved):

src/modules/claims/claims.service.ts
jrhender: @nichonien I think having expiry in milliseconds for `roleEIP191Jwt` is not correct. JWTs are required to have their dates as the number of seconds since Epoch, not milliseconds. https://stackoverflow.com/a/39926886 Is this how you understand it as well?

JGiter: Did-registry expects that expiration is in millyseconds. It converts expiration to seconds before signing https://github.com/energywebfoundation/ew-did-registry/blob/59183c7c1d013b2dc9f9f852b2af1915fa741eec/packages/jwt/src/JWTEIP191.ts#L75

JGiter: Probably we need to mention this in did-reg docs

jrhender: @JGiter but if we use `exp` in the credential payload directly instead of using `expirationTimestamp` (which I think we implemented in this PR https://github.com/energywebfoundation/ew-did-registry/pull/449/files), then I think the conversion from milliseconds to seconds that you linked to isn't happening, right?

JGiter: @jrhender Right, `expirationTimestamp` is set in `JwtPayload`, but `jwt` reads it from `JwtSignOptions`. Should `expirationTimestamp` be set in `JwtSignOptions`?

PR #654, opened 2022-09-12
Fix: publish assets on-chain

whitneypurdum:
### Description

Fix bug for missing subjectAgreement when publishing assets on-chain

### Contributor checklist


src/modules/claims/claims.service.ts
```diff
@@ -545,12 +545,16 @@ export class ClaimsService {
       return;
     }
 
-    if (!subjectAgreement && subject === this._signerService.did) {
+    if (!subjectAgreement) {
+      // If the signer is the subject, use the signer DID. If the signer is not the subject, use the subject DID:
+      const subjectAgreer =
+        subject === this._signerService.did ? this._signerService.did : subject;
       subjectAgreement = await this.approveRolePublishing({
-        subject: this._signerService.did,
```
whitneypurdum, line 550: @jrhender Thanks for the input on this. Is this what you were thinking? There is no concept in the claim of this being an asset, other than that the subject is different than signer. Is this logic what you were thinking? This causes an error in Claim Manager - see test case above (VM Exception while processing transaction: revert ClaimManager: agreement signer is not authorized to sign on behalf of subject), but maybe what you were getting at is in either case, the subject needs to be the signer service (although I don't think this is what you were getting at, as it wouldn't be true to the fact that the signer is not the subject of the claim).

e2e/claims.service.e2e.ts
```diff
@@ -948,6 +963,36 @@ describe('Сlaim tests', () => {
...
+          onChainProof: claim.onChainProof,
+          claimTypeVersion: version,
+          acceptedBy: claim.acceptedBy,
+          subject: myAssetDid,
+        };
+        mockGetClaimsBySubject
+          .mockReset()
+          .mockImplementationOnce(() => [mockedClaim]);
+
+        await claimsService.publishPublicClaim({
```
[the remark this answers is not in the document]
jrhender: @whitneypurdum from what I can tell, in the test case, the asset DID is a random end account and not an asset smart contract. This is the code that gives me this impression:
```
const myAsset = Wallet.createRandom().connect(provider);
const myAssetDid = `did:${Methods.Erc1056}:${Chain.VOLTA}:${myAsset.address}`;
```

Am I reading this right? I think that the test needs to actually use the asset contracts so that the contracts are there to represent the ownership relationship

src/modules/claims/claims.service.ts
whitneypurdum: @jrhender would it be safe to say that the subject is always the subject? I think the logic check is unnecessary, right? If the signer is the subject, or its not and you use the subject, you are always just using the subject. So it would be:

```
    if (!subjectAgreement) {
      // If the signer is the subject, use the signer DID. If the signer is not the subject, use the subject DID:
      subjectAgreement = await this.approveRolePublishing({
        subject: subject,
        role: claimType,
        version: +claimTypeVersion,
      });
      // }
    }
```
Is this right?

whitneypurdum: Is this right though, in the case of assets, can an asset be a subject that is agreeing to something, if it is not a real DID with signing capabilities?

jrhender: > There is no concept in the claim of this being an asset, other than that the subject is different than signer.

Yes, this is right

> would it be safe to say that the subject is always the subject?

Yes, I think so

> Is this right though, in the case of assets, can an asset be a subject that is agreeing to something, if it is not a real DID with signing capabilities

Well, the owner can update the DID document to add a key that can sign on behalf of the asset DID

jrhender (approved):
Looks good, thanks @whitneypurdum

PR #662, opened 2022-09-21
fix: bump ew-credentials to 2.2.1-alpha.296.0

whitneypurdum:
### Description

- Bumps ew-credentials version. If a user has a request to approve a role that she does not have the correct approval role for, the error will specify the role that she does not have. 
- Old Error: No authorative credential found for issuer
- New Error: No authorative credential found for issuer for role "installer.role..."

@jrhender and @JGiter It sounded from the [story](https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?assignee=608057a2e3b59800688dbc2a&selectedIssue=SWTCH-1455) that you wanted to add more info to this error. Should ICL catch the error from verifyIssuer, and if the error contains this text return it with more descriptive error message such as 'No authorative credential found for issuer for role "installer.role...". You must obtain and publish this role before approving this request.' 
Or should we just throw this new error? 

Update - we will just return this error and in the future address the root issue that SSI hub returns role approval requests that a user is not authorized to approve due to not having/publishing role

### Contributor checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,19 +1,19 @@
 {
   "name": "iam-client-lib",
-  "version": "6.2.1-alpha.2",
+  "version": "7.0.0-alpha.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "iam-client-lib",
-      "version": "6.2.1-alpha.2",
+      "version": "7.0.0-alpha.1",
       "license": "GPL-3.0-or-later",
       "dependencies": {
-        "@energyweb/credential-governance": "^2.2.1-alpha.295.0",
+        "@energyweb/credential-governance": "^2.2.1-alpha.296.0",
         "@energyweb/ekc": "^0.6.7",
-        "@energyweb/onchain-claims": "^2.2.1-alpha.295.0",
+        "@energyweb/onchain-claims": "^2.2.1-alpha.296.0",
         "@energyweb/staking-pool": "^1.0.0-rc.14",
-        "@energyweb/vc-verification": "^2.2.1-alpha.295.0",
+        "@energyweb/vc-verification": "^2.2.1-alpha.296.0",
         "@ensdomains/ens": "^0.6.2",
         "@ew-did-registry/claims": "0.7.1-alpha.816.0",
         "@ew-did-registry/credentials-interface": "0.7.1-alpha.816.0",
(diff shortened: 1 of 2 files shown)


JGiter:
> Should ICL catch the error from verifyIssuer, and if the error contains this text return it with more descriptive error message?
Yes, I would check if issuer has this role, but hasn't published it, then suggest to publish

whitneypurdum:
> > Should ICL catch the error from verifyIssuer, and if the error contains this text return it with more descriptive error message?
> > Yes, I would check if issuer has this role, but hasn't published it, then suggest to publish

@JGiter currently ew-credentials error pattern is to return 'boolean, string' to ICL. If we want to implement the above, we will need to do some string parsing to 1.) determine that this error is for not having authoritative credential and 2.) then pull out the actual role name. and then 3.) pass the role to check if the user has it. Unless there is a different way, or we want to change how we are handling errors. I feel we could achieve almost the same result by checking if the error is 'no authoritative credential found' and then adding "If you do not have this role, you must obtain it. If you already have this role, please publish it." But, happy to either update how we handle the errors (If we want to get the role and check for it, I suggest we change the ew-credential error to an object) or do regex parsing if we don't want to go this route but still want to check if the user has this role. Let me know if I'm missing something here. Thanks.

JGiter:
> currently ew-credentials error pattern is to return 'boolean, string' 
Do you mean error thrown during issuance? I think that the purpose of this task is to exlude such errors. So that when issuer receives request to issue, then issuance happens without any issues

JGiter:
My last comment doesn't correspond to actual task description which say about error handling, but nevertheless I think we should be move toward avoiding error

JGiter:
But yes, if error happend you can append it with message saying that issuer role needs to be published and rethrow

whitneypurdum:
> But yes, if error happend you can append it with message saying that issuer role needs to be published and rethrow

At this point we don't know if it needs to be published, or if he has not enrolled in the role at all. I think we should throw an error saying "No authorative credential found for issuer for role "installer.role...". If you do not have this role, you must obtain it. If you do have this role, please publish it."

Either way on SB, it will appear in the red error box. There is no other UI treatment right now for caught errors. That would be an addition to this task I think.

jrhender (approved):
Seems good to me

PR #663, opened 2022-10-11
style: add warn for shadow variables (variables in scope that share s…

whitneypurdum:
…ame name)

### Description
I wasn't involved in conversation about this, but thought I would give it a try.  (https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1437)

Extends eslint.rc to add warn for shadow variables https://eslint.org/docs/latest/rules/no-shadow

You can see some no-shadow warnings here https://github.com/energywebfoundation/iam-client-lib/pull/663/files, is this what you were intending to achieve with this rule? 
### Contributor checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   extends: ['@energyweb'],
   rules: {
-    "@typescript-eslint/no-shadow": "warn",
+    "@typescript-eslint/no-shadow": ["warn", {ignoreFunctionTypeParameterNameValueShadow: true}],
   },
   env: {
     browser: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   extends: ['@energyweb'],
   rules: {
-    "@typescript-eslint/no-shadow": ["warn", {"ignoreFunctionTypeParameterNameValueShadow": true}],
+    "@typescript-eslint/no-shadow": "warn",
   },
   env: {
     browser: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,8 @@
 module.exports = {
   extends: ['@energyweb'],
+  rules: {
+    "@typescript-eslint/no-shadow": "warn",
+  },
   env: {
     browser: true,
     es2021: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   extends: ['@energyweb'],
   rules: {
-    "@typescript-eslint/no-shadow": ["warn", {ignoreFunctionTypeParameterNameValueShadow: true}],
+    "@typescript-eslint/no-shadow": ["warn", {"ignoreFunctionTypeParameterNameValueShadow": true}],
   },
   env: {
     browser: true,
--- a/src/modules/domains/domains.service.ts
+++ b/src/modules/domains/domains.service.ts
@@ -462,7 +462,11 @@ export class DomainsService {
         );
       } else {
         for await (const { namespace: ns } of apps) {
-          await this.changeAppOwnership({ namespace: ns, newOwner, returnSteps });
+          await this.changeAppOwnership({
+            namespace: ns,
+            newOwner,
+            returnSteps,
+          });
         }
       }
     }
@@ -1414,6 +1418,6 @@ export class DomainsService {
   }
 
   get domainReader(): DomainReader {
-   return this._domainDefinitionReader;
+    return this._domainDefinitionReader;
   }
 }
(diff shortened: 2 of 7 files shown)


whitneypurdum:
I wasn't involved in conversation about this, but thought I would give it a try.  (https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1437)

JGiter (commented):
This rule isn't specific to iam-lib. Would it make sense to add it to shared config https://github.com/energywebfoundation/shared-configs/tree/master/packages/eslint-config?

jrhender:
@whitneypurdum would you mind fixing a handful of the new shadow warnings in this PR so that we can see what it looks like please 🙇‍♂️? No need to fix them all to start, but I think we should fix them before we merge this as we don't want to introduce a change which then leads to people seeing warnings when they are development. IMO, that isn't very pleasant dev experience.

src/modules/domains/domains.service.ts
```diff
@@ -473,18 +473,18 @@ export class DomainsService {
       );
     }
 
-    const steps = changeOwnerNamespaces.map((namespace) => {
```
whitneypurdum, line 476: @jrhender This is a fix for an error involving a map. I personally think its better to call a spade a spade and say what it is, rather than come up with different names for the argument, just to avoid repeating, but that's just my thought. What do you think? cc @JGiter

whitneypurdum: You can see below a lot of the warnings are for using 'namespace' in a map.

.eslintrc.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   extends: ['@energyweb'],
+  rules: {
+    "@typescript-eslint/no-shadow": ["warn", {"ignoreFunctionTypeParameterNameValueShadow": true}],
```
whitneypurdum, line 4: @jrhender @JGiter I added this ignoreFunctionTypeParameterNameValueShadow so that you didn't get a warning when you had a function param with a name that is repeated, so we can call parameters what they actually are rather than come up with different words for the same thing every time we have a method. Does this sound okay to you?

src/modules/domains/domains.service.ts
whitneypurdum: With the above change with ignoreFunctionTypeParameterNameValueShadow, we would not have to do this for every .map or every method with a param that has a shared name (like 'namespace')

JGiter: I think lenght of variable name should depend on size of variable scope. This `namespace` is used only for one operation - `changeDomainOwnerTx`, thus I would name it just by `n`, but variable occures several times in order to convert return type of `changeOrgOwnership` to `MulticallTx` (which is also array of `ReturnStepWithRetryCheck`). This conversion happens in other places too. So you can separate conversion from `tx` to `ReturnStepWithRetryCheck` and then the code in question would look like
```
 const steps = changeOwnerNamespaces.map((n) => {
      return this.txToCall(
         this.changeDomainOwnerTx({ newOwner, namespace: n }), 
         ()=>await this.getOwner({namespace: n}) === newOwner
   )
}

txToCall(tx, retryCheck) {...}
```
I would also change `getOwner({namespace})` to `getOwner(namespace)`. This is not breaking change, because `getOwner` is protected

.eslintrc.js
JGiter, line 4: Is this rule allows to shadow function arguments by local variables? If so I think it is largely breaks the purpose of the rule

whitneypurdum:
> @whitneypurdum would you mind fixing a handful of the new shadow warnings in this PR so that we can see what it looks like please 🙇‍♂️? No need to fix them all to start, but I think we should fix them before we merge this as we don't want to introduce a change which then leads to people seeing warnings when they are development. IMO, that isn't very pleasant dev experience.

@jrhender I fixed a few, there are 19 no-shadow warnings split between e2e and non-e2e files. What is the goal of this PR - to fix all instances of no-shadow? Or just in non e2e?

src/modules/domains/domains.service.ts
jrhender: I'm also not the biggest fan of single letter variable names, but I think they have their place in loops, closures, in-line functions. I did a bit of SO research and found this post which is saying that single letter variable names are reasonable within a lexical scope of about 25 lines, which seems about right to me. https://stackoverflow.com/a/5803019 
So I think the single letter solution could work here. What do you think @whitneypurdum ?

whitneypurdum: Works for me! What is "SO"?

whitneypurdum: Oh, stackoverflow.

jrhender (approved):
@whitneypurdum 
> I fixed a few, there are 19 no-shadow warnings split between e2e and non-e2e files. What is the goal of this PR - to fix all instances of no-shadow? Or just in non e2e?

IMO, if we are enabling the rule, then we should resolve all warnings that result from it (as you've done, thank you)
I think there is just one more to fix on line 1062 of `claims.service.e2e.ts` and then this is good to merge, IMO

PR #675, opened 2023-02-10
task/SWTCH-2020 get access tokens from cookies

JGiter:
### Description

For security reasons access tokens should not be sent in response body. To fix this ssi-hub will send tokens in cookies. This PR allows Node.js client to extract tokens from cookies
Since `CacheServerClient.authenticate` is getting large, setting of tokens is refactored

### Contributor checklist

### DIFF
--- a/e2e/auth.e2e.ts
+++ b/e2e/auth.e2e.ts
@@ -116,9 +116,6 @@ describe('Authentication tests', () => {
       expect(
         cacheClient['_httpClient'].defaults.headers.common.Authorization
       ).toBe(`Bearer ${newTokens.token}`);
-      expect(cacheClient['refresh_token']).toStrictEqual(
-        newTokens.refreshToken
-      );
       expect(nockScope.isDone()).toBe(true);
     });
 
--- a/e2e/auth.e2e.ts
+++ b/e2e/auth.e2e.ts
@@ -109,8 +109,15 @@ describe('Authentication tests', () => {
 
       cacheClient['refresh_token'] = oldRefreshToken;
 
-      await expect(cacheClient['refreshToken']()).resolves.toStrictEqual(
-        newTokens
+      await cacheClient['refreshToken']();
+      expect(cacheClient['refresh_token']).toStrictEqual(
+        newTokens.refreshToken
+      );
+      expect(
+        cacheClient['_httpClient'].defaults.headers.common.Authorization
+      ).toBe(`Bearer ${newTokens.token}`);
+      expect(cacheClient['refresh_token']).toStrictEqual(
+        newTokens.refreshToken
       );
       expect(nockScope.isDone()).toBe(true);
     });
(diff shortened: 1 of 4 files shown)


src/modules/cache-client/cache-client.service.ts
```diff
@@ -72,43 +73,26 @@ export class CacheClient implements ICacheClient {
...
     // First try to refresh access token
     try {
-      const refreshedTokens = await this.refreshToken();
-      tokens = refreshedTokens;
-      setTokens();
-
-      if (!tokens || !(await this.isAuthenticated())) {
-        tokens = undefined;
-      }
+      await this.refreshToken();
```
artursudnik, line 78: why not log error details here? We very often suffer because of no details when debugging

src/modules/cache-client/cache-client.service.ts
```diff
@@ -681,18 +665,48 @@ export class CacheClient implements ICacheClient {
...
+        refreshToken = refreshTokenCookie.value;
+      }
+    }
+    if (!token || !refreshToken) {
+      token = data.token;
+      refreshToken = data.refreshToken;
+    }
+    this.refresh_token = refreshToken;
+    this._httpClient.defaults.headers.common[
+      'Authorization'
```
artursudnik, line 709: so the server is going to expect a token in the Authorization header, not the actual cookie? What about the browser then?

e2e/auth.e2e.ts
```diff
@@ -109,8 +109,15 @@ describe('Authentication tests', () => {
...
-      await expect(cacheClient['refreshToken']()).resolves.toStrictEqual(
-        newTokens
+      await cacheClient['refreshToken']();
+      expect(cacheClient['refresh_token']).toStrictEqual(
+        newTokens.refreshToken
+      );
+      expect(
+        cacheClient['_httpClient'].defaults.headers.common.Authorization
+      ).toBe(`Bearer ${newTokens.token}`);
+      expect(cacheClient['refresh_token']).toStrictEqual(
```
artursudnik, line 119: is this repeated on purpose?

src/modules/cache-client/cache-client.service.ts
JGiter: Ok, could be usefull

e2e/auth.e2e.ts
JGiter: Thanks, will fix

src/modules/cache-client/cache-client.service.ts
JGiter: Server expects token based on client. When iam lib runs in Node.js this method sets token in header, when iam runs in browser this method isn't called and cookies managed by browser

artursudnik: I see. So, the server needs to expect token coming from both `Authorization` and `Cookie` headers. What if both are sent and worse if they are different because of some bug?

artursudnik: OK. So, the `Authorization` header takes precedence over Cookie now. 👍

src/modules/cache-client/cache-client.service.ts
```diff
@@ -681,18 +667,48 @@ export class CacheClient implements ICacheClient {
...
+
+      if (tokenCookie && refreshTokenCookie) {
+        token = tokenCookie.value;
+        refreshToken = refreshTokenCookie.value;
+      }
+    }
+    if (!token || !refreshToken) {
+      token = data.token;
+      refreshToken = data.refreshToken;
+    }
```
jrhender, line 708: What's the rationale for these lines? It seems to be covering the case where the tokens aren't returned in the cookie headers. Is this right? I guess this could be useful to have but just trying to understand a bit better 😄

jrhender (approved):
Looks good to me, thanks @JGiter

src/modules/cache-client/cache-client.service.ts
JGiter: I left this temporarily until we will change ssi-hub to send tokens only in cookies

PR #678, opened 2023-02-15
task/SWTCH-2036 siwe

JGiter:
### Contributor checklist

- [x] Tests - add new or update existed test for changes you made.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,7 +6,7 @@ module.exports = {
     },
     transformIgnorePatterns: ["node_modules/!(@energyweb/ekc)", "node_modules/.+\\.!(mjs)$"],
     testEnvironment: "node",
-    testRegex: "(/(e2e|src)/(.|\\.)*\\.(e2e|spec)\\.ts$)",
+    testRegex: "(/(e2e|src)/(.|\\.)*auth\\.(e2e|spec)\\.ts$)",
     moduleFileExtensions: ["ts", "tsx", "js", "json"],
     coveragePathIgnorePatterns: ["/node_modules/", "/e2e/", "/test/"],
     coverageThreshold: {
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,7 +6,7 @@ module.exports = {
     },
     transformIgnorePatterns: ["node_modules/!(@energyweb/ekc)", "node_modules/.+\\.!(mjs)$"],
     testEnvironment: "node",
-    testRegex: "(/(e2e|src)/(.|\\.)*auth\\.(e2e|spec)\\.ts$)",
+    testRegex: "(/(e2e|src)/(.|\\.)*\\.(e2e|spec)\\.ts$)",
     moduleFileExtensions: ["ts", "tsx", "js", "json"],
     coveragePathIgnorePatterns: ["/node_modules/", "/e2e/", "/test/"],
     coverageThreshold: {
--- a/src/config/cache.config.ts
+++ b/src/config/cache.config.ts
@@ -1,14 +1,18 @@
 import { EWC_CHAIN_ID, VOLTA_CHAIN_ID } from '@energyweb/credential-governance';
 import { CacheServerClientOptions } from '../modules/cache-client/cache-client.types';
 
-const cacheConfig: Record<number, CacheServerClientOptions> = {
+const defaultConfig: Record<number, CacheServerClientOptions> = {
   [VOLTA_CHAIN_ID]: {
     url: 'https://identitycache-staging.energyweb.org/v1/',
     cacheServerSupportsAuth: true,
+    auth: {
+      domain: 'https://switchboard-dev.energyweb.org',
+    },
   },
   [EWC_CHAIN_ID]: {
     url: 'https://identitycache.energyweb.org/v1/',
     cacheServerSupportsAuth: true,
+    auth: { domain: 'https://switchboard.energyweb.org' },
   },
 };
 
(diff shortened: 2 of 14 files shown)


src/modules/cache-client/cache-client.service.ts
```diff
@@ -85,16 +86,10 @@ export class CacheClient implements ICacheClient {
     // If refresh token failed or access token is not valid, then sign new identity token
     if (!(await this.isAuthenticated())) {
       getLogger().info('[CACHE CLIENT] obtaining new tokens');
-      delete this._httpClient.defaults.headers.common['Authorization'];
```
jrhender, line 88: Is this line still needed? It seems to be unrelated to the authentication protocol change

jrhender (commented):
This is looking good 👍

src/modules/cache-client/cache-client.service.ts
JGiter: I removed it, because this line do nothing. Next step in this method signs in to ssi-hub, which doesn't need Authorization header and then Authrization header is assigned new value. But of course we can remove this in separate PR

jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     },
     transformIgnorePatterns: ["node_modules/!(@energyweb/ekc)", "node_modules/.+\\.!(mjs)$"],
     testEnvironment: "node",
-    testRegex: "(/(e2e|src)/(.|\\.)*\\.(e2e|spec)\\.ts$)",
+    testRegex: "(/(e2e|src)/(.|\\.)*auth\\.(e2e|spec)\\.ts$)",
```
jrhender, line 9: I think we should remove this before merging, right?

src/config/cache.config.ts
```diff
@@ -1,14 +1,18 @@
 import { EWC_CHAIN_ID, VOLTA_CHAIN_ID } from '@energyweb/credential-governance';
 import { CacheServerClientOptions } from '../modules/cache-client/cache-client.types';
 
-const cacheConfig: Record<number, CacheServerClientOptions> = {
+const defaultConfig: Record<number, CacheServerClientOptions> = {
   [VOLTA_CHAIN_ID]: {
     url: 'https://identitycache-staging.energyweb.org/v1/',
```
jrhender, line 6: If default domain is `switchboard-dev`, we should change `url` to `identitycache-dev`. I think this is actually a good time to do this as this is already a breaking change due to support for the previous login method being removed.

```suggestion
    url: 'https://identitycache-dev.energyweb.org/v1/',
```

src/modules/cache-client/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+   * Signs in to ssi-hub. Authentication method configurations are looked for in `cacheConfigs`
+   *
+   * @returns authentication tokens
+   */
+  async signIn(): Promise<AxiosResponse<AuthTokens>> {
+    const {
+      data: { nonce },
+    } = await this.http.post<{ nonce: string }>('/login/siwe/initiate');
+    const siweMessage = new SiweMessage({
+      statement: `I accept the Terms of Service: ${this.config.domain}`,
```
jrhender, line 33: I would remove the `statement` for now, unless it is necessary.
```suggestion
```

src/modules/cache-client/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+      nonce,
+      domain: new URL(this.config.domain).host,
+      address: this.signerService.address,
+      uri: `${cacheConfigs()[this.signerService.chainId].url}login/siwe/verify`,
```
jrhender, line 37: I think this is potentially a bit brittle as it will have a missing`/` if `url` doesn't end with a `/`. I'm not sure what the better approach is though. Maybe using something like this? 
```suggestion
      const url = cacheConfigs()[this.signerService.chainId].url
      uri: `${url}${url.substring(url.length - 1) === '/' ? '' : '/'} login/siwe/verify`,
```

src/config/cache.config.ts
jrhender: @JGiter could you also note the above in the migration guide documentation please? If folks had previously relied on the default value, they will need to specify `'https://identitycache-staging.energyweb.org/v1/' as the default has been changed to 'https://identitycache-dev.energyweb.org/v1/'

JGiter: Good idea

src/modules/cache-client/auth.ts
JGiter: Ok

src/modules/cache-client/auth.ts
JGiter: Yes, it is a bit dangerous. I changed it to new URL('/login/siwe/verify', cacheConfigs()[this.signerService.chainId].url).href. This will create correct uri regardless of trailing slash

nichonien (approved):
Thanks @JGiter , great work!

jrhender:
@JGiter Can we move https://github.com/energywebfoundation/iam-client-lib/pull/678/commits/e004d1a1f4291650297e9e0d3397e6a1d26df06a to a separate PR?
I think the PR as it was prior to the commit was mergeable (IMO) and the changes introduced by the commit might require further discussion

JGiter:
> @JGiter Can we move [e004d1a](https://github.com/energywebfoundation/iam-client-lib/commit/e004d1a1f4291650297e9e0d3397e6a1d26df06a) to a separate PR? I think the PR as it was prior to the commit was mergeable (IMO) and the changes introduced by the commit might require further discussion

Ok, I moved this change to another branch if we decide to proceed with such approach https://github.com/energywebfoundation/iam-client-lib/tree/sign_non_eip-191

PR #679, opened 2023-02-23
dd sign non eip-191

JGiter:
### Contributor checklist

### DIFF
--- a/src/modules/signer/signer.service.ts
+++ b/src/modules/signer/signer.service.ts
@@ -1,5 +1,4 @@
 import { BigNumber, providers, utils, Wallet } from 'ethers';
-import base64url from 'base64url';
 import {
   TypedDataDomain,
   TypedDataField,
@@ -13,7 +12,6 @@ import {
   executionEnvironment,
 } from '../../utils/detect-environment';
 import {
-  IPubKeyAndIdentityToken,
   ProviderType,
   ProviderEvent,
   AccountInfo,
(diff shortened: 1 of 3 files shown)


src/modules/signer/signer.service.ts
```diff
@@ -354,6 +345,19 @@ export class SignerService {
...
+   * @param message Stringified message
+   */
+  async sign(message: string): Promise<string> {
+    const messageBytes = Buffer.from(message);
+    const signature = await this.signer.signMessage(messageBytes);
+    if (!this._publicKey || this._isEthSigner === undefined) {
+      await this._determineSigner(messageBytes, signature);
+    }
+    return signature;
+  }
```
jrhender, line 359: I don't get the "this method does not guaraties that signature will conform to EIP-191" as `signer.signMessage` performs an EIP-191 signature, right? 

https://docs.ethers.org/v5/api/signer/#Signer-signMessage
> A signed message is prefixd with "\x19Ethereum Signed Message:\n" and the length of the message, using the [hashMessage](https://docs.ethers.org/v5/api/utils/hashing/#utils-hashMessage) method, so that it is [EIP-191](https://eips.ethereum.org/EIPS/eip-191) compliant. If recovering the address in Solidity, this prefix will be required to create a matching hash.

JGiter: I think this is requirements to implementers of `Signer` interface. But not all implementations are conform to this. If I remember correctly this is the reason why we are prefixing message manually before passing it to Metamask signer https://github.com/energywebfoundation/iam-client-lib/blob/9d01b84fb1d9dd67971688b010811ad786050538/src/modules/signer/signer.service.ts#L347

PR #681, opened 2023-03-02
task/SWTCH-2114 allow different ssi-hub and siwe uri

JGiter:
### Contributor checklist

### DIFF
--- a/src/modules/cache-client/cache-client.types.ts
+++ b/src/modules/cache-client/cache-client.types.ts
@@ -1,4 +1,3 @@
-import { SiweMessage } from 'siwe';
 import { AssetHistoryEventType } from '../assets';
 
 export enum Order {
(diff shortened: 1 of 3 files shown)


src/modules/cache-client/cache-client.types.ts
```diff
@@ -1,12 +1,11 @@
-import { SiweMessage } from 'siwe';
 import { AssetHistoryEventType } from '../assets';
 
 export enum Order {
   'ASC' = 'ASC',
   'DESC' = 'DESC',
 }
 
-export type SiweOptions = Pick<SiweMessage, 'domain'>;
+export type SiweOptions = { domain: string; uri?: string };
```
jrhender, line 8: When we have a moment, we should add `uri` and `domain` to the example in the README, IMO https://github.com/energywebfoundation/iam-client-lib#2-connect-to-the-cache-server

jrhender (approved):

PR #683, opened 2023-04-02
feat/retry request to auth service

JGiter:
### Description

- Retry handler is moved from cache service to auth service
- Retry hanlder changed from `promise-retry` to `axios-retry`, which allowed to skip `makeRetryRequest` wrapers around each cache client request
- Tests refactored avoiding mocking of implementation wherever possible and mocking API instead 

### Contributor checklist


src/modules/cache-client/cache-client.service.ts
```diff
@@ -82,210 +78,172 @@ export class CacheClient implements ICacheClient {
   }
 
   async getRoleDefinition(namespace: string) {
-    return await this.makeRetryRequest(async () => {
-      const { data } = await this._httpClient.get<IRole>(`/role/${namespace}`);
-      return data?.definition;
-    });
+    const { data } = await this._httpClient.get<IRole>(`/role/${namespace}`);
+    console.dir(data, { depth: 3, colors: true });
```
jrhender, line 82: It would be better to use `getLogger` and use the logger here if we would like to log the result.

JGiter: sorry, I missed that after debugging

jrhender (approved):

nichonien (approved):

PR #684, opened 2023-04-14
Replace bili with rollup

JGiter:
### Description

Currently use `bili` build tool is poorly maintained. I suppose this is the reason of failing to update dependencies. This PR replaces `bili` with `rollup`. In addition of better tool it also provides simpler configuration 

### Contributor checklist


package.json
```diff
@@ -64,18 +63,18 @@
...
-    "@energyweb/vc-verification": "2.2.1-alpha.313.0",
+    "@energyweb/vc-verification": "2.2.1-alpha.314.0",
     "@ensdomains/ens": "^0.6.2",
     "@ew-did-registry/claims": "0.8.1-alpha.1037.0",
     "@ew-did-registry/credentials-interface": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did-document": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did-ethr-resolver": "0.8.1-alpha.1037.0",
-    "@ew-did-registry/did-ipfs-store": "0.7.1-alpha.816.0",
+    "@ew-did-registry/did-ipfs-store": "0.8.0",
```
jrhender, line 77: I'm not sure this should be done as doesn't think upgrade to the ENS cluster version?

JGiter:
> This PR looks like a good start but `lib.esm` and `lib` should be added to `.gitignore` I believe

Yes, build artifacts are happened in root by mistake. I'll output build to dist. And no changes will be needed in .gitignore, because `dist` already ignored

JGiter:
This build tested with Switchboard and no issues were found. The only required change is to replace internal imports like [here](https://github.com/energywebfoundation/switchboard-dapp/blob/9708cc04f23aaca4f7e8b22751d8cc2a3609497e/src/app/shared/services/signer-facade/signer-facade.service.ts#L3) with `iam-client-lib/dist/lib.esm/src/modules/signer/signer.types`

jrhender:
> The only required change is to replace internal imports like [here](https://github.com/energywebfoundation/switchboard-dapp/blob/9708cc04f23aaca4f7e8b22751d8cc2a3609497e/src/app/shared/services/signer-facade/signer-facade.service.ts#L3) with iam-client-lib/dist/lib.esm/src/modules/signer/signer.types

Do you think this means that we would need to mark this as a breaking change? On one hand, it is breaking because updating the new version would require code changes and on the other, the code changes would be references to internal file locations which I would say we are free to change.

In the PR description, you wrote:
> esm and cjs builds are put in different folders lib and lib.esm, which is common practice

Do you have any documentation/evidence for this being a common or best practice?

JGiter:
Formally it is not a breaking change because there are neither API nor functional changes. Client code needs to be updated because of improper use of library.
Hmm, may be I exaggerated by saying that it is common practice. I just saw such separation in several popular libriries, like [ethers](https://github.com/ethers-io/ethers.js/blob/main/package.json) or **[web3](https://github.com/web3/web3.js/blob/4.x/packages/web3-core/package.json)**

package.json
```diff
@@ -2,20 +2,19 @@
   "name": "iam-client-lib",
   "version": "8.0.0-alpha.4",
   "description": "Library for Decentralized Identity and Access Management",
-  "main": "dist/index.js",
-  "module": "dist/index.esm.js",
-  "types": "dist/src/index.d.ts",
+  "main": "dist/lib/index.js",
+  "module": "dist/lib.esm/index.js",
+  "types": "dist/lib/src/index.d.ts",
```
jrhender, line 7: I am more convinced that `lib` is the right directory to put files that would be consumed by a package manager like `npm` 👍 . `dist` seems to be the directory for a bundled file for webpage or CDNs to load directly.
However, given this, I think that the `lib` directory should not be under the `dist` directory. I think `lib` should be at the top level.
```suggestion
  "main": "lib/index.js",
  "module": "lib.esm/index.js",
  "types": "lib/src/index.d.ts",
```

JGiter:
Something happend to GH. `git status` says that local is up-to-date with remote, while local is ahead of several commits. Will close and reopen this PR

JGiter:
Something happend to GH. `git status` says that local is up-to-date with remote, while local is ahead of several commits. Will close and reopen this PR

PR #686, opened 2023-05-09
Build with rollup

JGiter:
### Description

Replacing bili with rollup, because bili is not maintained anymore, which prevents from fixing

### Contributor checklist


tsconfig.test.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["e2e/**/*"],
+  "ts-node": {
+    "compilerOptions": { "module": "CommonJS" }
+  }
```
jrhender, line 6: Is this needed? The root `tsconfig.json` from which this config extends already has these `ts-node` options, I believe.

tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["src/**/*", "ethers/**/*"],
+  "ts-node": {
+    "compilerOptions": { "module": "CommonJS" }
+  }
```
jrhender, line 6: Is this needed? The root `tsconfig.json` from which this config extends already has these `ts-node` options, I believe.

package.json
```diff
@@ -64,18 +63,18 @@
...
-    "@energyweb/vc-verification": "2.2.1-alpha.313.0",
+    "@energyweb/vc-verification": "2.2.1-alpha.314.0",
     "@ensdomains/ens": "^0.6.2",
     "@ew-did-registry/claims": "0.8.1-alpha.1037.0",
     "@ew-did-registry/credentials-interface": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did-document": "0.8.1-alpha.1037.0",
     "@ew-did-registry/did-ethr-resolver": "0.8.1-alpha.1037.0",
-    "@ew-did-registry/did-ipfs-store": "0.7.1-alpha.816.0",
+    "@ew-did-registry/did-ipfs-store": "^0.7.1-alpha.816.0",
```
jrhender, line 77: Why the addition of the caret here? I think a pinned version is clearly and maybe a bit safer (and the other 
`@ew-did-registry` packages have a pinned version)

package.json
```diff
@@ -2,20 +2,19 @@
...
+  "types": "lib/src/index.d.ts",
   "files": [
-    "dist"
+    "lib",
+    "lib.esm"
   ],
   "scripts": {
     "lint": "eslint ./src/**/*.ts ./e2e/**/*.ts",
-    "prebuild": "rimraf dist ethers",
-    "build": "npm run build:contracts && npm run build:ts && copyfiles \"ethers/**/*.d.ts\" dist",
```
jrhender, line 14: Is `copyfiles \"ethers/**/*.d.ts\" dist` no longer needed? If it is not needed anymore, perhaps `copyfiles` can be removed as a dev dependency

jrhender (commented):
A few comments/questions but looks like a good change 👍 . Thanks for your work on this @JGiter and apologies for the slow PR review

tsconfig.build.json
JGiter: yes, it can be removed :+1:

package.json
JGiter: yes, contract definitions are part of the code, so they are included in build. I will remove `copyfiles`. good catch. may be we can use https://www.npmjs.com/package/npm-check

nichonien (approved):

jrhender (approved):
Looks good to me, thanks @JGiter

PR #687, opened 2023-06-05
Ignore renovate did-ipfs-store

JGiter:
### Contributor checklist


nichonien (approved):
Thanks @JGiter , this was a much needed. Would be a good Idea to do in other repos as well.

PR #692, opened 2023-06-06
Update siwe@2.1.4

JGiter:
### Description

Build https://github.com/energywebfoundation/iam-client-lib/actions/runs/5187681109/jobs/9350646272 fails, because peer dependency @octokit/retry-error of @octokit/plugin-retry is not installed. This peer dependency ignored because of `legacy-peer-deps` npm flag, which was added in order to use ethers@5.7 for siwe@1. Now when latest siwe depends on ethers@5.7 we can disable legacy peer

### Contributor checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-legacy-peer-deps = true
+# legacy-peer-deps = true
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "nats.ws": "^1.7.1",
     "qs": "^6.9.4",
     "set-cookie-parser": "^2.5.1",
-    "siwe": "^1.1.6",
+    "siwe": "^2.1.4",
     "ts-interface-checker": "^1.0.2",
     "tslib": "^2.0.3",
     "uuid": "^7.0.3"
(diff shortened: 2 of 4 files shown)


.npmrc
jrhender, line 1: Could we simply remove the whole file rather than commenting out the line?

JGiter: yes, sure

package.json
```diff
@@ -102,7 +102,7 @@
     "nats.ws": "^1.7.1",
     "qs": "^6.9.4",
     "set-cookie-parser": "^2.5.1",
-    "siwe": "^1.1.6",
+    "siwe": "^2.1.4",
```
jrhender, line 105: Question for you @JGiter . As this is a major version upgrade, are there any breaking changes here? If so, do we need to update `siwe` in passport-did-auth/did-auth-proxy in parallel so that the SIWE messages created by iam-client-lib are accepted by the backend apps? Or will messages created with v2 siwe be accepted by servers running v1 siwe?

JGiter: I didn't notice breaking changes. `siwe` v2 already used in ssi-hub and [passport-did-auth](https://github.com/energywebfoundation/passport-did-auth/blob/7bc46205872e090d96fe0d7254f359bc81365726/package.json#L68). I also tested ICL with `siwe` 2.1.3 locally and everything went well. Though there is something happened in 2.1.4, so I am downgrading siwe to 2.1.3

JGiter: Actually problem with 2.1.4 was for some temporal reason. So I'll bring 2.1.4 back

PR #698, opened 2023-06-12
Bundle tslib

JGiter:
### Description

This is temporal fix for issue with `rollup`. The issue is that `rollup` importing `tslib` as local dependency like this
```
import { __awaiter } from '/home/runner/work/iam-client-lib/iam-client-lib/node_modules/tslib/tslib.es6.js'
```
This PR temporarily excludes `tslib` from external dependencies and bundles it

We also does not directly import `tslib`, so there is no need to add to package.json https://github.com/energywebfoundation/iam-client-lib/blob/9df32764c2d5036a5be2d82ecfa7c916c7386582/package.json#L107

### Contributor checklist

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -18,7 +18,7 @@ module.exports = [
         tsconfig: 'tsconfig.build.json',
       }),
     ],
-    external: [/node_modules/],
+    external: [/node_modules(?!\/tslib)/],
   },
   {
     input: 'src/index.ts',
@@ -36,6 +36,6 @@ module.exports = [
         tsconfig: 'tsconfig.build.json',
       }),
     ],
-    external: [/node_modules/],
+    external: [/node_modules(?!\/tslib)/],
   },
 ];


jrhender:
@JGiter Thanks for the description

> This is temporal fix for issue with rollup. The issue is that rollup importing tslib as local dependency like this
> `import { __awaiter } from '/home/runner/work/iam-client-lib/iam-client-lib/node_modules/tslib/tslib.es6.js'`

What do you mean by "temporal fix"? Do you mean that this a fix that we will want to revert at some point?

Could you explain a bit more why the local dependency import that you've shown is a problem? I'm not saying it's not, it's just not obvious to me why it is.

JGiter:
That's right, `tslib` is external in fact and a proper fix is to make rollup to bundle it as `import { __awaiter } from 'rollup'`

jrhender:
> With import from '/home/runner/work/iam-client-lib/iam-client-lib/node_modules/tslib/tslib.es6.js'  npm will look tslib on any machine in /home/runner...

Ah thanks, I can see now why that's a problem

> a proper fix is to make rollup to bundle it as `import { __awaiter } from 'rollup'`

How would this fix work? This fix looks to me like `__awaiter` is somehow bundled in `rollup`.

JGiter:
> How would this fix work? This fix looks to me like `__awaiter` is somehow bundled in `rollup`.

I beleive that `import from 'rollup'` just instructs npm to resolve `rollup` from npm.js when `iam-client-lib` is installed

jrhender:
Ok, I'm still trying to understand how the tslib import works, but I think that you're point about ts-lib in package.json might not be correct:

> We also does not directly import tslib, so there is no need to add to package.json

From rollup TS plugin docs: https://github.com/rollup/plugins/tree/master/packages/typescript#install

> Note that both typescript and tslib are peer dependencies of this plugin that need to be installed separately.

jrhender:
Wouldn't the ideal import statement in bundled code be `import { __awaiter } from 'tslib';` not `import { __awaiter } from 'rollup';` ?

JGiter:
> Wouldn't the ideal import statement in bundled code be `import { __awaiter } from 'tslib';` not `import { __awaiter } from 'rollup';` ?

Yes, sorry for confusion. I meant `import from 'tslib`

jrhender (approved):
I think this isn't the ideal fix but I think this is worth a try, once we test locally that this works.

JGiter:
Strange, but the issue did not go away. I linked ICL locally to switchboard, but still `npm start` throws
```
./node_modules/iam-client-lib/lib.esm/index.js:1:0-131 - Error: Module not found: Error: Can't resolve '/home/runner/work/iam-client-lib/iam-client-lib/node_modules/tslib/tslib.es6.js' in './switchboard-dapp/node_modules/iam-client-lib/lib.esm'
```
though lib.esm/index.js does not import tslib

PR #761, opened 2024-07-12
bug/swtch-3035 Not valid VC are verified

JGiter:
### Description

Checks issuer verification result

### DIFF
--- a/src/modules/claims/claims.service.ts
+++ b/src/modules/claims/claims.service.ts
@@ -76,6 +76,7 @@ import {
   RoleEIP191JWT,
   isEIP191Jwt,
   IssuerResolver,
+  VerificationResult,
 } from '@energyweb/vc-verification';
 import { DidRegistry } from '../did-registry/did-registry.service';
 import { ClaimData, isClaimService } from '../did-registry/did.types';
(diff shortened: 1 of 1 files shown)


src/modules/claims/claims.service.ts
```diff
@@ -1482,11 +1483,20 @@ export class ClaimsService {
...
+          role
+        );
       } else {
-        await this._issuerVerification.verifyIssuer(issuerDID.id, role);
+        issuerVerificationResult = await this._issuerVerification.verifyIssuer(
+          issuerDID.id,
+          role
+        );
       }
+      issuerVerified = issuerVerified && issuerVerificationResult.verified;
```
JGiter, line 1498: `issuerVerified` is also used as result of status verification, so the final `issuerVerified` is boolean `AND` of both checks. Perhaps this checks should be separated

JGiter:
The CI is not passing, but it is something wrong with setting up ipfs as locally tests are passing if I run them separately. The exception is `claims.service.e2e`. The tests are never completes locally for me

src/modules/claims/claims.service.ts
```diff
@@ -1482,10 +1483,22 @@ export class ClaimsService {
...
-        await this._issuerVerification.verifyIssuer(issuerDID.id, role);
+        issuerVerificationResult = await this._issuerVerification.verifyIssuer(
+          issuerDID.id,
+          role
+        );
+      }
+      const { verified, error } = issuerVerificationResult;
+      issuerVerified = issuerVerified && verified;
+      if (error.length > 0) {
+        errors.push(error);
```
jrhender, line 1501: `push` probably pushes the array and not the elements of the array. Can either use `concat` or spread operator

JGiter: @jrhender It turns out that `error` is a string https://github.com/energywebfoundation/ew-credentials/blob/c9f5790277d09c5ea5c332e15ee15add880f6e6d/packages/vc-verification/src/models/credentials-types.ts#L11

jrhender: Ah okay. I think that `!== ""` is a more idomatic way of checking for an empty string. Would reduce this string/array confusion when reading the code.
```suggestion
      if (error !== "") {
        errors.push(error);
```

jrhender (approved):
A small suggestion, but seems good to me

PR #762, opened 2024-07-16
Set issuer resolver

JGiter:


whitneypurdum (approved):

src/modules/claims/claims.service.ts
```diff
@@ -109,7 +110,7 @@ export class ClaimsService {
   private _claimManagerInterface = ClaimManager__factory.createInterface();
   private _claimRevocation: ClaimRevocation;
   private _issuerVerification: IssuerVerification;
-  private _issuerResolver: EthersProviderIssuerResolver;
+  private _issuerResolver?: IssuerResolver;
```
jrhender, line 113: @JGiter why this change? Is there a code path wherein `_issuerResolver` is undefined

JGiter: I changed meaning of this field. Now this is overriding resolver, which might not be set. I will add comment. May be also rename it somehow?

jrhender: So you're saying this code wouldn't work? From my reading of the code it should be fine. I'm testing it out locally....
```suggestion
  private _issuerResolver: IssuerResolver;
```

JGiter: Generally typescript should error this when [strictProperyInitialization](https://www.typescriptlang.org/tsconfig/#strictPropertyInitialization) is true. In our [configuration](https://github.com/energywebfoundation/iam-client-lib/blob/880ae80a2dcebd8c2d1974bc28b2c59207ff48fe/tsconfig.json#L35) it is false, but error is not thrown even if to change it to true. May be this rule overwritten somewhere

JGiter: Actually typescript indeed reject not initialized _issuerResolver if enable `strictPropertyInitialization`

jrhender: ah okay, I see. Good point about `strictPropertyInitialization`. However, as I think we currently have `strictPropertyInitialization = false` and as this would affect all of the other private properties in this class (and several other of the service classes), I'm not sure it makes sense to deviate from the pattern here. Not a big deal though. I think we can proceed either way 👍

jrhender (approved):
Thanks @JGiter , I think we should proceed 👍

PR #763, opened 2024-07-16
Skipping test from `deploy` action

JGiter:
Commit type is `fix` in order to trigger release

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -46,6 +46,7 @@ jobs:
         run: npm run build
 
       # Skipped, until fix issue with setting up IPFS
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -45,8 +45,9 @@ jobs:
       - name: Build package
         run: npm run build
 
-      - name: Npm Run Test
-        run: npm run test
+      # Skipped, until fix issue with setting up IPFS
```
jrhender, line 48:
```suggestion
      # Skipped, until fix issue with setting up IPFS
      # See https://energyweb.atlassian.net/browse/SWTCH-3073
```

jrhender (approved):

PR #765, opened 2024-07-18
Publish v8.1.1

JGiter:


whitneypurdum (approved):

whitneypurdum:
@JGiter this is for the issuer where if the verification of issuer is "false" it was not throwing error, right?

JGiter:
> @JGiter this is for the issuer where if the verification of issuer is "false" it was not throwing error, right?

Exactly

PR #766, opened 2024-07-25
bug/retry authenticate

JGiter:
Refreshing access token in axios-retry callback is not reflected in axios configuration used for initial failed request, so retry requests will continue to fail even though ICL logs that user authenticated


artursudnik (approved):

docs/api/classes/modules_assets.AssetsService.md
JGiter, line 1: don't really know why so many changes in the documentation. May be `typedoc` was updated

